Issue #7, opened 2015-12-16
bot.say does not work with # prefixed channel name


metoikos:
When you try to send a message to channel with # prefixed channel name nothing is happening. 
```
bot.say({
    text: "Hello World!",
    channel: "#random"
});
```
Instead of # prefixed channel name, when you try with channel id everything works great.

```
bot.say({
    text: "Hello World!",
    channel: 'C0XXXXXXX'
});
```

ben-stover:
Please try it again but with single quotes like: '#random' instead of double quotes - but keep text double quoted. Thanks

ben-stover:
Has your bot been allowed into the # prefixed channel?

tonythomson:
I believe this is a documentation error; `bot.say()` uses the RTM API, which expects a channel ID.

ZuzooVn:
@metoikos What's your development environment?

metoikos:
Hello @ZuzooVn 
OS X El Capitan

```bash
➜  ~  uname -a
Darwin rocko 15.0.0 Darwin Kernel Version 15.0.0: Sat Sep 19 15:53:46 PDT 2015; root:xnu-3247.10.11~1/RELEASE_X86_64 x86_64
➜  ~  node --version
v4.2.2
```

ZuzooVn:
I can not build the source code with 

`➜  ~  uname -a
Darwin hostname.local 15.2.0 Darwin Kernel Version 15.2.0: Fri Nov 13 19:56:56 PST 2015; root:xnu-3248.20.55~2/RELEASE_X86_64 x86_64`

`➜  ~   node --version
v5.0.0`

metoikos:
I installed with `npm install botkit` command.
It generates some errors but it's not related to botkit, it is bufferutil issue. You can check out at issue https://github.com/howdyai/botkit/issues/2

anonrig:
@metoikos How do you learn the channel id?

tonythomson:
We've drifted off-topic from the issue... @benbrown Any plans for a chatroom/forum/email list?

@anonrig A list of current channel IDs is available from the [channel list API](https://api.slack.com/methods/channels.list), but it's probably better to just use the data you get from the bot's initial connection to RTM, and then keep that list up to date. You can edit the file `Slackbot_worker.js` (see line 75 or so) to have the bot cache different data sets for the Slack team to which you're connecting (e.g. users, channels, IMs, etc.)

benbrown:
[...]

PR #8, opened 2015-12-17
Fix param for  bot.say() example code in readme.md

tonythomson:
Assuming issue #7 is a documentation error, this updates the code example for `bot.say()` in the readme with an accurate example value.
The docs on [this page](http://howdy.ai/botkit/docs/#sending-messages) will need the same fix.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -222,7 +222,7 @@ for more information about listening for and responding to messages.
 
 It is also possible to bind event handlers directly to any of the enormous number of native Slack events, as well as a handful of custom events emitted by Botkit.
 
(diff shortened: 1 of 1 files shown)


montanaflynn:
Would also be nice if the docs showed a way to get the channel id, or if the botkit had a helper method to do so.

RafaelCosman:
@montanaflynn the channel id comes with every message. Do you mean that you want a function from #channel names to channel ids?

montanaflynn:
@rafaelcosman yes that is what I meant

RafaelCosman:
@tonythomson I just tested this to be sure. You're completely correct.
@benbrown I'm merging this in.

PR #11, opened 2015-12-17
don't expect to merge this but want some feedback

brianleroux:
Adding some tests to help understand the API and document workings. To run them you'll need a `.env` file in the root (which will be `.gitignored`) that has the following:

```
TOKEN=ur-slack-token-here
```
You can get a token from the bottom of https://api.slack.com/web

From there you just run `npm test` to run the tests. (Don't forget to `npm i to get the dev deps.) Things work as you'd expect…but I'm not sure how to close an RTM session. Any help appreciated!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ examples/db_slackbutton_slashcommand/
 examples/db_team_bot/
 .DS_Store
 */.DS_Store
+.env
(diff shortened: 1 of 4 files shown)


brianleroux:
opening a new PR w/ a fix!

PR #12, opened 2015-12-17
Support multiple message types

zemaj:
This enables support for multiple message types.

For example, the following will now work;
```
bot.reply(message, {"type": "typing"});

setTimeout(function() {
    bot.reply(message, "Hello there!");
}, 2000);
```
which will show the bot as typing before sending a response. This is particularly useful when you have to perform some computation before replying after a few seconds.


benbrown:
Nicely done.

PR #14, opened 2015-12-17
Add 'yep' to yes utterances

maxehmookau:
Let's be honest, this is about as small a PR as could possibly exist, but I use `yep` a lot, as do lots of people. (Maybe it's a British thing?)


benbrown:
Fabulous!! Thanks.

PR #16, opened 2015-12-17
Fixed bug that made sample bot crash when not being able to write an emoji

MiguelSR:
Fixed bug that made sample bot crash when not being able to write an emoji


benbrown:
Thank you!

PR #19, opened 2015-12-18
adds tests and a fix for clean exit

brianleroux:
- adds tape to start testing out the api and harden the code
- adds tap-spec for nice reporting
- fixes a bug on closeRTM for clean exit

### DIFF
--- a/test/Slack_web_api-test.js
+++ b/test/Slack_web_api-test.js
@@ -1,0 +1,39 @@
+var test = require('tape')
+var Botkit = require('../')
+var env = require('node-env-file')
+var path = require('path')
+
+env(path.join(__dirname, '..', '.env'))
+var token = {token:process.env.TOKEN}
+
+test('sanity', t=> {
+  t.plan(4)
+  t.ok(token, '.env sets TOKEN')
+  console.log(token)
+  t.ok(Botkit, 'Botkit exists')
+  t.ok(Botkit.core, 'Botkit.core exists')
+  t.ok(Botkit.slackbot, 'Botkit.slackbot exists')
+  console.log(Botkit)
+})
+
+test('can start and then stop a bot', t=> {
+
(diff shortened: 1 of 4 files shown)


test/Slack_web_api-test.js
```diff
@@ -0,0 +1,39 @@
...
+  t.ok(Botkit, 'Botkit exists')
+  t.ok(Botkit.core, 'Botkit.core exists')
+  t.ok(Botkit.slackbot, 'Botkit.slackbot exists')
+  console.log(Botkit)
+})
+
+test('can start and then stop a bot', t=> {
+
+  var controller = Botkit.slackbot({debug:false}) 
+  var bot = controller.spawn(token).startRTM((err, bot, payload)=> {
```
anonrig: If you're using arrow functions, why not use const or let in all of your code?

RafaelCosman:
@brianleroux thanks for this pull request. Sorry for taking so long to look at it. I'll test it shortly.

benbrown:
As I discussed with @brianleroux, we just need to make a small change to the way the interval is being cleared before we can merge this to avoid breaking the slack button bot functionality where multiple bots are running in one process...

benbrown:
OK! I made the change I wanted and merged this in!

Woo, tests!!

RafaelCosman:
Thanks for looking at this, @benbrown!

anonrig:
If this pull request is merged, botkit will not work on Nodejs  < 4.0.0

benbrown:
@anonrig only the tests are > 4 though right?

benbrown:
@anonrig I just refactored the tests to be more friendly to earlier Nodes.

Let me know if you have other concerns!

anonrig:
Yes only the tests were Node 4+ supported. Right now it seems fine. But in the future it is important to switch to NodeJS 4+. At least the LTS version 4.2.3 @benbrown. What do you think?

Issue #27, opened 2015-12-20
Add option to pass `redirect_uri`


lambtron:
https://api.slack.com/docs/oauth

For some reason when I was going through the Add to button example, I kept running into issues around passing the redirect. Even though it says its optional in Slack's documentation, after I added the `redirect_uri` (in two places—1. in the [Slack bot config](https://github.com/howdyai/botkit/blob/master/lib/SlackBot.js#L27) and 2. in the [oauth function](https://github.com/howdyai/botkit/blob/master/lib/SlackBot.js#L266)).

PR #28, opened 2015-12-20
added option to pass redirect_uri

lambtron:
Addresses this issue https://github.com/howdyai/botkit/issues/27


RafaelCosman:
@lambtron #35 is clearly a duplicate of this. Seems like naming is the biggest difference. Do you have any strong opinions about if we should merge this one or #35?

benbrown:
Let's go with this one - though the is one place in the docs that will need to be updated to match the new redirectUri parameter name.

PR #30, opened 2015-12-20
added mongo storage client

lambtron:
since this uses [`monk`](https://github.com/Automattic/monk), there are some known bugs that are commonly encountered if the node version is 4 or above.

Notably: 
https://github.com/Automattic/monk/issues/91
https://github.com/Automattic/monk/issues/114

I chose Monk for its simplicity, though Mongoose is probably the safer bet if people want to use newer node versions (i'm using 0.12 for this currently).

Just an FYI!

### DIFF
--- a/lib/mongo_storage.js
+++ b/lib/mongo_storage.js
@@ -2,12 +2,11 @@ var db = require('monk');
 
 module.exports = function(config) {
 
-  if (!config) throw new Error('Need to provide mongo address.');
+  if (!config && !config.mongo_uri) throw new Error('Need to provide mongo address.');
 
-  var teams_db = db(path).get('teams');
-  var users_db = db(path).get('users');
-  var channels_db = db(path).get('channels');
-
+  var teams_db = db(config.mongo_uri).get('teams');
+  var users_db = db(config.mongo_uri).get('users');
+  var channels_db = db(config.mongo_uri).get('channels');
 
   var storage = {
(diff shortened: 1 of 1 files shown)


lib/mongo_storage.js
```diff
@@ -0,0 +1,49 @@
+var db = require('monk');
+
+module.exports = function(path) {
```
RafaelCosman: Nice PR, @lambtron. In order to provide a consistent interface, this should be ```function(options)``` where options is an object with a path property.

lambtron: Oh shoot, good point ;)

I can make the update.

audibleblink:
I love that Monk returns promises when you don't give it a callback. I hadn't heard of it, so I'm glad I saw this. I liked it so much that I went off and made [my own simple adapter](https://github.com/audibleblink/casbot2016/blob/develop/lib/mongo_storage.js) (I don't intend to hijack your PR, and I'm focusing on Harmony anyway.) My use-case needed more than the standard 3 stores.

lambtron:
@audibleblink great to see you found use from Monk :)

lib/mongo_storage.js
RafaelCosman: Thanks @lambtron! I'm reviewing and testing this now and we should have it merged soon. Thanks for your patience.

PR #33, opened 2015-12-20
Implements a Firebase storage system

RafaelCosman:


RafaelCosman:
This pull request needs to pass ```storage_test.js``` before it can be considered.

lib/firebase_storage.js
```diff
@@ -0,0 +1,95 @@
+/*
```
anonrig: What do you think about JSDoc? http://usejsdoc.org/

RafaelCosman: @anonrig looks useful! I assume that you don't mean just for this file but for the whole codebase, right?

RafaelCosman: If that's the case, would you please either submit a pull request or an issue? Thanks!

anonrig: Thats the case. Will submit a pull request!

PR #35, opened 2015-12-21
Custom OAuth clientRedirectUri handling

guillaumepotier:
As seen in `#27`, I did not find a way to pass a custom redirect uri (my app have many) to Botkit. This small PR add this capability.


guillaumepotier:
Damn. Did not see #28 and made a duplicate :(

RafaelCosman:
Yep, this is a duplicate of #28. Seems like naming is the biggest difference. Do you have any strong opinions about if we should merge this one or #28?

Issue #37, opened 2015-12-21
Redis storage


guillaumepotier:
Hi,

I wrote a simple Redis storage using redis `hashes` storage to handle data. Are you interested by it? Do you want me to make a PR here and add it somewhere in the repo, or create a dedicated repo and link it in the README?

Best

RafaelCosman:
So far people (including me) have been submitting PRs for these, but it's not clear if that's what the managers of the repo want. If they accept our PRs today then we'll know the answer.

benbrown:
Yes! Please submit a pull request!! That is a great feature to have!

guillaumepotier:
Okay @benbrown, I just submitted my PR in #38, feel free to comment. I [factorized my code](https://github.com/howdyai/botkit/pull/38/files#diff-efceb781e59d7290103716d0c7b77fa3R21) to handle the API that do not change much bw `teams`, `users` and `channels` and used a code that could be simpler with more lines. If not enough readable I could develop. What I particularly like with this implementation is that very easily [you could add sets](https://github.com/howdyai/botkit/pull/38/files#diff-efceb781e59d7290103716d0c7b77fa3R18), I use that to store some of my application business (tokens or sessions to be able to call my API).

RafaelCosman:
@guillaumepotier I left a note on your PR about passing ```storage_tests.js``` before we merge it.

guillaumepotier:
Since my second pull request have been merged, I think we could close this issue.

Best

PR #38, opened 2015-12-21
Added simple Redis storage

guillaumepotier:
Hi there,

Following #37 here is a humble submission for a simple Redis storage.

I'm just starting with botkit (and thanks a lot) and I already know I cannot use file storage since my bots would live in various VMs.

I use node redis (`npm install redis --save`) and [hashes](http://redis.io/commands#hash) sets (that have O(1) complexity like `SET` and `GET` and is a lot easier for `all()` method O(N) on the set vs. O(N) on the entire database for `KEYS`).

This is not tested in production (yet) and should not make any BC Break (just moved `simple_storage.js` into subdir if some other storages are going to be in the repo).

Tried to update a bit the README with my french background..

Please let me know if some things are wrong or missing, won't change my life if not merged, but could be nice for a broader adoption of the lib.

Best


RafaelCosman:
@guillaumepotier thanks very much for making this pull request! Can you make this storage system pass ```storage_test.js```? I just wrote that so all of our storage systems will be tested and interchangeable. Thanks!

guillaumepotier:
Hi,

great to have some tests. I'm looking at them right now. I noticed thanks to the tests that `all()` methods returns an array. I find that not so handy since we deal with objects and users/teams/channels ids everywhere, my redis storage was returning an object like that : 

```javascript
{
  "id1": {"id": "id1", "foo": "bar"},
  "id2": {"id": "id2", "bar": "baz"}
}
```

I'll update my `all()` method than and maybe could I implement (for myself and the sake of others pushing the curiosity looking at the code) a `allById()`.

PR #39, opened 2015-12-21
Inject RegExp match into message

guillaumepotier:
Another PR here! :p

I used to code on Hubot before trying Botkit and lack a feature I did not found (and maybe missed?) which is accessing the data matching the capturing `RegExp` in the `hears()` function, like [here in Hubot](https://hubot.github.com/docs/scripting/#capturing-data)

This PR inject the `match` object into `message` to deal with it more easily in the callback. Maybe I'll add some documentation in the README.

Best

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -640,14 +640,16 @@ function Botkit(configuration) {
       events = events.split(/\,/g);
     }
 
+    var match;
     for (var k = 0; k < keywords.length; k++) {
       var keyword = keywords[k];
       for (var e = 0; e < events.length; e++) {
         (function(keyword) {
           botkit.on(events[e],function(bot,message) {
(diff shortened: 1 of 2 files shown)


RafaelCosman:
@guillaumepotier thanks for this pull request. I agree that the matches should be passed with the message because I've had to regenerate them myself sometimes which is clearly wasteful. The match that comes back might be confusing for some users because it looks like this:

```[ 'open the front doors',
  'front',
  index: 0,
  input: 'blah open the front doors blah' ]```

guillaumepotier:
Hi,

Yes this could be disturbing for someone not accustomed to RegExp, but sometimes (often?) I have multiple capturing groups and having the whole object could be handy :)

Thanks for the merge btw!

Best

PR #40, opened 2015-12-22
fixes slash_command and outgoing webhook functions

audibleblink:
This PR addresses the `botkit is undefined` error when attempting to implement custom slash-commands. Changing it to `slack_botkit` allows it to work again and I'm assuming that's what was meant to be there.

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -89,7 +89,7 @@ function Slackbot(configuration) {
 
             res.status(200);
 
-            var bot = botkit.spawn(team);
+            var bot = slack_botkit.spawn(team);
 
             bot.team_info = team;
             bot.res = res;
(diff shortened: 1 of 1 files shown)


audibleblink:
Also @benbrown , I wanted to start working on the FIX THIS comments in `Slackbot.js`, but with out a bit more context I'm not sure which direction to take. 

Why does `#createWebhookEndpoints` assume storage at all? 

I can get slash-commands to work if I add a `controller.storage.teams.save({REDACTED})` right after controller spawning but I don't think it should be necessary. What do you think?

benbrown:
@audibleblink It assumes storage because our test cases were built assuming the use of the slack button.  

It *shouldn't* be required though, and maybe we can adjust it so that if the slack app configuration hasn't been specified, it doesn't try to look up the team in the DB.

benbrown:
Also thanks for the PR!

PR #42, opened 2015-12-24
fix incorrect variable when checking id

willrax:
Noticed that this checks for the presence of a user id rather than a channel id. :)


benbrown:
Thank you!

PR #44, opened 2015-12-27
Package dependencies updated.

okvic77:
There were some errors while installing on node 4 and 5.

I update the dependencies and problem gone.

This site could help to be up-to-date. [https://www.bithound.io]()


RafaelCosman:
@okvic77 Thanks for this pull request! I just tested this with ```demo_bot.js``` and it works fine. I'll merge it in.

PR #64, opened 2016-01-02
Add loglevels and an extension point for an external logging provider

zetlen:
- Added a configuration option `logger`, which expects an object with a `log` method. 
   - The `log` method expects its first argument to be either a string or an integer representing one of the RFC 5424 syslog levels.
   - The `log` method accepts a variable number of additional arguments and should try to inspect and log all of them.

 - If a `logger` is not supplied to `botkit.slackbot()`, then the default implementation, `ConsoleLogger`, will be used.
  - `ConsoleLogger` implements much the same behavior as before, with the addition of log levels.
  - `ConsoleLogger` will honor a configuration option `logLevel` supplied to `botkit.slackbot()`. It can be either a string or an integer representing one of the RFC 5424 syslog levels. Any message with a higher log level than the `logLevel` will not be logged.

- Added helper methods to `botkit.log` for each of the RFC 5424 syslog levels. They call the configured `logger` with the appropriate log level as first argument. They are:
  - `botkit.log.emergency()`
  - `botkit.log.alert()`
  - `botkit.log.critical()`
  - `botkit.log.error()`
  - `botkit.log.warning()`
  - `botkit.log.notice()`
  - `botkit.log.info()`
  - `botkit.log.debug()`

- Went through existing codebase and replaced log messages that seemed to clearly be errors, warnings, and notices with calls to the appropriate method. No lower log levels seemed to be called for (such as `emergency`, `alert`, or `critical`).

- The existing methods `botkit.log()` and `botkit.debug()`, and their behavior, are preserved, with the addition of log levels (they are now aliases to `botkit.log.info()` and `botkit.log.debug()`, respectively).

- The existing configuration options `log === false` and `debug`, and their behavior, are preserved and applied to `ConsoleLogger`. They are translated into `logLevel` configurations (`error` and `debug`, respectively). If a `logLevel` configuration is also supplied, then it will override these options.

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -78,7 +78,7 @@ function Slackbot(configuration) {
           // this won't work for single team bots because the team info
           // might not be in a db
           if (err || !team) {
-            slack_botkit.logerror('Received slash command, but could not load team');
+            slack_botkit.log.error('Received slash command, but could not load team');
 
           } else {
             message.type='slash_command';
@@ -118,7 +118,7 @@ function Slackbot(configuration) {
           // this won't work for single team bots because the team info
           // might not be in a db
           if (err || !team) {
-            slack_botkit.logerror('Received outgoing webhook but could not load team');
+            slack_botkit.log.error('Received outgoing webhook but could not load team');
           } else {
             message.type='outgoing_webhook';
 
@@ -351,7 +351,7 @@ function Slackbot(configuration) {
 
                   slack_botkit.saveTeam(team,function(err,id) {
                     if (err) {
-                      slack_botkit.logerror('An error occurred while saving a team: ',err);
+                      slack_botkit.log.error('An error occurred while saving a team: ',err);
                       if (callback) {
                         callback(err,req,res);
                       } else {
@@ -380,7 +380,7 @@ function Slackbot(configuration) {
                         slack_botkit.storage.users.save(user,function(err,id) {
 
                           if (err) {
-                            slack_botkit.logerror('An error occurred while saving a user: ',err);
+                            slack_botkit.log.error('An error occurred while saving a user: ',err);
                             if (callback) {
                               callback(err,req,res);
                             } else {
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -78,7 +78,7 @@ function Slackbot(configuration) {
           // this won't work for single team bots because the team info
           // might not be in a db
           if (err || !team) {
-            slack_botkit.log('Received slash command, but could not load team');
+            slack_botkit.logerror('Received slash command, but could not load team');
 
           } else {
             message.type='slash_command';
@@ -118,7 +118,7 @@ function Slackbot(configuration) {
           // this won't work for single team bots because the team info
           // might not be in a db
           if (err || !team) {
-            slack_botkit.log('Received outgoing webhook but could not load team');
+            slack_botkit.logerror('Received outgoing webhook but could not load team');
           } else {
             message.type='outgoing_webhook';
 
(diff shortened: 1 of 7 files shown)


lib/SlackBot.js
```diff
@@ -78,7 +78,7 @@ function Slackbot(configuration) {
           // this won't work for single team bots because the team info
           // might not be in a db
           if (err || !team) {
-            slack_botkit.log('Received slash command, but could not load team');
+            slack_botkit.logerror('Received slash command, but could not load team');
```
anonrig: What do you think about making a `log` object that has an error function like `slack_botkit.log.error` which is more readable.

RafaelCosman:
@zetlen Thanks for this pull request and the very detailed description! Could you take a look at @anonrig's comment? I'm fine with either ```slack_botkit.logerror``` or ```slack_botkit.log.error``` as he recommends. Let us know which you prefer to do and then I'll merge this in.

Thanks very much!

-- @rafaelcosman

anonrig:
Even if we are going to use `logerror`, I think it would be better to use camel case while defining function names.

For example:
Instad of `botkit.logerror` => `botkit.logError`

zetlen:
That's a fair point, @anonrig. Latest commit changes those methods to be dot lookup: `slack_botkit.logerror` has become `slack_botkit.log.error`.

Thanks @RafaelCosman!

lib/SlackBot.js
zetlen: I did this. `slack_botkit.log` is still a function, but now it has methods for each severity as well.

anonrig:
Thank you @zetlen 
Additionally `This branch has conflicts that must be resolved`

RafaelCosman:
@zetlen if you resolve those we'll merge this in! Thanks for your work on this. :ship:

zetlen:
Yuck, just my luck to open a pull request right before a big formatting change gets merged. :expressionless:

anonrig:
Sorry about that @zetlen :(

lib/Slackbot_worker.js
```diff
@@ -27,11 +27,11 @@ module.exports = function(botkit, config) {
             botkit.debug('CANNOT SEND WEBHOOK!!');
 
             return cb && cb('No webhook url specified');
-        }
+  }
 
         request.post(bot.config.incoming_webhook.url, function(err, res, body) {
-            if (err) {
-                botkit.debug('WEBHOOK ERROR',err);
+          if (err) {
```
anonrig: You should use `editorconfig` and `.jscsrc` for styling. Indentation is wrong :(

lib/Slackbot_worker.js
```diff
@@ -39,31 +39,31 @@ module.exports = function(botkit, config) {
         }).form({ payload: JSON.stringify(options) });
     };
 
-    bot.configureRTM = function(config) {
-        bot.config.token = config.token;
-        return bot;
+  bot.configureRTM = function(config) {
```
anonrig: Can you also remove these lines? It breaks the indentation.

lib/Slackbot_worker.js
zetlen: Aha! I'll fix those. How are you running `jscs`? When I run `jscs ./lib` I get hundreds of errors. Can I suggest adding `"pretest": "jscs ./lib"` to `npm scripts` so others don't make the mistake I made?

anonrig: I am not familiar with it adding pretest command. You sound you know what you're doing :+1: 

Can you add it to package.json with this pull request? Thank you.

zetlen:
OK, all merged. I'm gonna open a separate pull request for incorporating style checking into test. :1234:

lib/Slackbot_worker.js
zetlen: Thought it should be a separate pull request for good bookkeeping, but I can merge it into this one if you prefer. Your call.

PR #73, opened 2016-01-03
add reply with typing + start typing

mikehmorrissey:
I felt that the bot's replies were pretty jarring / abrupt. I think introducing a slight delay and adding a typing indicator will provide a much smoother and natural user experience. 

I added a method to reply with typing delay based on how long the message is. I also added a method to simply start typing which is useful when calling out to third party apis. 

Let me know what you think.

### DIFF
--- a/lib/Slackbot_worker.js
+++ b/lib/Slackbot_worker.js
@@ -358,6 +358,38 @@ module.exports = function(botkit, config) {
         bot.say(msg,cb);
     };
 
+    /**
+     * sends a typing message to the source channel
+     * @param {Object} message source
+     */
+    bot.startTyping = function(src) {
+        bot.reply(src, { type: 'typing' })
+    }
+
+    /**
+     * replies with message after typing delay
+     * @param {Object} message source
+     * @param {(string|Object)} response string or object
+     * @param {function} optional request callback
+     */
+    bot.replyWithTyping = function(src, resp, cb) {
+        var text;
+
+        if (typeof(resp) == 'string') {
+            text = resp;
+        } else {
+            text = resp.text;
+        }
+
+        var typingLength = 1200 / 60 * text.length;
+        typingLength = typingLength > 2000 ? 2000 : typingLength;
+
+        bot.startTyping(src)
+
+        setTimeout(function() {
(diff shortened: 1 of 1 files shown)


lib/Slackbot_worker.js
```diff
@@ -358,6 +358,28 @@ module.exports = function(botkit, config) {
         bot.say(msg,cb);
     };
 
+    bot.startTyping = function(src) {
```
anonrig: We started using Javascript Documentation found in http://usejsdoc.org.

Can you update all of your functions with necessary documentations?

lib/Slackbot_worker.js
```diff
@@ -358,6 +358,28 @@ module.exports = function(botkit, config) {
...
+    bot.replyWithTyping = function(src, resp, cb) {
+        var text;
+
+        if (typeof(resp) == 'string') {
+            text = resp;
+        } else {
+            text = resp.text;
+        }
+
+        var typingLength = 500 / 60 * text.length;
```
anonrig: You've assumed every letter takes 8.33 miliseconds to type. For a sentence of 10 characters it takes 83.33 ms which is faster then most of the humans. In order to make it more human noticable, I suggest you increase your typingLength value.

lib/Slackbot_worker.js
mikehmorrissey: @anonrig will do :+1:!

mikehmorrissey:
@anonrig thanks for the feedback! made the changes.

mateeyow:
@lunarmayor Is it possible to add the `typing` feature for converations like @MattieTK said.

kevinsuh:
yes, this has not been added to conversations. would love to have a convo.sayWithTyping option

maxhartshorn:
I agree this would be a natural fit into conversations

mvaragnat:
I may be doing something wrong, but the "is typing" indicator doesn't disappear when the message is posted. it continues for a couple of seconds (before timing out ? I'm not sure).

devhaoy:
Did anyone manage to make typing available in convo?  Pretty sure every one making a messenger bot needs this features

PR #74, opened 2016-01-05
fix issue where events would come in as an array

gizm0duck:
running slackbutton_bot.js would crash on startup because events were coming in as an array already.

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -640,7 +640,11 @@ function Botkit(configuration) {
 
     botkit.on = function(event, cb) {
         botkit.debug('Setting up a handler for',event);
-        var events = event.split(/\,/g);
+        if (typeof(event) == 'string')
+          var events = event.split(/\,/g);
+        else
+          var events = event;
+
         for (var e in events) {
             if (!this.events[events[e]]) {
                 this.events[events[e]] = [];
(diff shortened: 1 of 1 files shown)


lib/CoreBot.js
```diff
@@ -640,7 +640,11 @@ function Botkit(configuration) {
 
     botkit.on = function(event, cb) {
         botkit.debug('Setting up a handler for',event);
-        var events = event.split(/\,/g);
+        if (typeof(event) == 'string')
```
anonrig: Can you put this line into a single if statement?

gizm0duck: Just put onto a single line, let me know if that's acceptable.

RafaelCosman:
Good catch, @gizm0duck! I think that I introduced this bug. It's good that `hears()` and `on()` will now both support both syntaxes.

Testing your PR now...

RafaelCosman:
Just made this pass the linter: `5 passing (1s)`.

RafaelCosman:
This works great. Thanks for fixing this, @gizm0duck. Merging now.

lib/CoreBot.js
```diff
@@ -640,7 +640,8 @@ function Botkit(configuration) {
 
     botkit.on = function(event, cb) {
         botkit.debug('Setting up a handler for',event);
-        var events = event.split(/\,/g);
+          var events = (typeof(event) == 'string') ? event.split(/\,/g) : event;
```
anonrig: Can you use `isArray` function? `Array.isArray(event)` is more readable.

Issue #75, opened 2016-01-05
"Getting Started" instructions are missing a step


benbrown:
The getting started instructions are missing a step instructing the user to run npm install to install all necessary dependencies.

Also, it should be made clear that those instructions assume the reader has cloned the repo from Github, not installed it via npm.

PR #78, opened 2016-01-05
Add npm i instructions to readme

Glowbal:
Installation in readme was missing instructions for installing the dependencies through `npm i`.

#75


anonrig:
Can you add the difference for `npm install --production` and the normal `npm install`

RafaelCosman:
@Glowbal, thanks for responding to @anonrig's comment. Merging this now.

PR #79, opened 2016-01-05
Adds a react() function to Conversations

RafaelCosman:
```
convo.react("+1");
```

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -127,6 +127,12 @@ function Botkit(configuration) {
             this.addMessage(message);
         };
         
+        /**
+         * Reacts to the user's most recent messsage.
+         * @param reaction the reaction to react with. For example, "+1".
+         * For documentation on the API call see https://api.slack.com/methods/reactions.add
+         * @author @rafaelcosman
+         */
         this.react = function(reaction) {
             bot.api.reactions.add({
                 timestamp: this.transcript[this.transcript.length-1].ts,
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -126,16 +126,17 @@ function Botkit(configuration) {
         this.say = function(message) {
             this.addMessage(message);
         };
-        
+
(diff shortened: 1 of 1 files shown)


lib/CoreBot.js
```diff
@@ -126,6 +126,18 @@ function Botkit(configuration) {
         this.say = function(message) {
             this.addMessage(message);
         };
+        
+        this.react = function(reaction) {
```
anonrig: Can you add a JSDoc?

lib/CoreBot.js
```diff
@@ -126,6 +126,18 @@ function Botkit(configuration) {
...
+            bot.api.reactions.add({
+                timestamp: this.transcript[this.transcript.length-1].ts,
+                channel: this.source_message.channel,
+                name: reaction
+            },function(err, res) {
```
anonrig: Did you use the linter? 

According to linter there should be an empty space between `,` and `function`

Please test your code using `npm test`

RafaelCosman:
@anonrig Thanks! 

Also, I think that while this approach can work, it's not ultimately the right way to do it. Nothing in `CoreBot.js` should be calling the Slack API. Those calls should be in one of the slack-specific files.

RafaelCosman:
Now passes linter. `5 passing (1s)`. Still would not recommend merge because this code should really be abstracted out of `CoreBot.js`.

anonrig:
What about making the reaction an `enum`? @RafaelCosman

karlbright:
+1 to moving this out of Corebot before it's merged and the idea of making this an `enum` is pretty sweet.

RafaelCosman:
@anonrig and @karlbright, great suggestions! I have another PR that I'm focusing on right now so I'm going to temporarily mark this PR with #DONT_MERGE. I'll leave it open so that others who want to work on this or just want to use reactions can use this temporarily until we get a complete fix.

PR #84, opened 2016-01-07
Instruct how to get dependencies w/ git install

tswast:
People who aren't 100% familiar with npm might not realize how easy it is to install the dependencies after checking out directly from git.

For example, I got the error: `Error: Cannot find module 'mustache'` when I first tried running botkit as described in the Getting Started section.


tswast:
Oh! I see there's even more thorough instructions pending with https://github.com/howdyai/botkit/pull/78. I'll close this PR.

PR #85, opened 2016-01-07
fix reply with typing linter errors

mikehmorrissey:
@benbrown didn't mean to introduce these in the first place but here's some fixes to get the linter passing in slackbot worker


RafaelCosman:
@mikehmorrissey, testing this now...

RafaelCosman:
Ok, `5 passing (1s)`. Merging this now.

PR #86, opened 2016-01-07
add middleware entry point to preprocess messages

mikehmorrissey:
@benbrown i've been using this in my fork of botkit and I think at lot of people would find it useful. it introduces a middleware layer before triggering the direct_message, ambient, direction_mention and mention events. This allows for message normalization/preprocessing. for example I could write something like this: 

```js
const normalize = (bot, message, next) => {
  message.text = message.text.replace('gonna', 'going to')
  next()
}

const BotController = Botkit.slackbot({
  // can be an array of functions
  middleware: normalize
})
```

that's a pretty simple example but i can imagine people writing middleware to normalize common typos, slang, contractions, etc. What do you think? I'll write up the documentation and provide a solid example if you're into it.

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -1,5 +1,6 @@
 var Botkit = require(__dirname + '/CoreBot.js');
 var request = require('request');
+var ware = require('ware');
 var express = require('express');
 var bodyParser = require('body-parser');
 
@@ -37,6 +38,10 @@ function Slackbot(configuration) {
         return slack_botkit;
 
     };
+
+    // add preprocessing middleware
+    slack_botkit.preprocessor = ware();
+    slack_botkit.preprocessor.use(configuration.preprocessor || []);
 
     // set up a web route that is a landing page
     slack_botkit.createHomepageEndpoint = function(webserver) {
@@ -417,7 +422,6 @@ function Slackbot(configuration) {
         slack_botkit.log('** Setting up custom handlers for processing Slack messages');
         slack_botkit.on('message_received',function(bot, message) {
 
-
             if (message.ok != undefined) {
                 // this is a confirmation of something we sent.
                 return false;
@@ -469,7 +473,10 @@ function Slackbot(configuration) {
 
                     message.event = 'direct_message';
 
-                    slack_botkit.trigger('direct_message',[bot,message]);
+                    slack_botkit.preprocessor.run(bot, message, function(err, bot, message) {
+                        slack_botkit.trigger(message.event, [bot, message]);
+                    });
+
                     return false;
 
                 } else {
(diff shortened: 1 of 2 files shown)


RafaelCosman:
@mikehmorrissey this seems useful! I think it would be great to have a set of standard normalization functions that we could drop in. Let's reach out in dev4slack.slack.com and see if anyone's interested in collaborating on this. Also let's poll the level of interest in this feature. I know that I have at least one bot that should be using something like this.

DEGoodmanWilson:
It would be nice—bear in mind I'm somewhat new to Node and Node idioms—if things were namespaces in such a way as to be able to add new middleware at different points in the flow. For example, I am wanting to add a middleware hook around local cacheing of channel/user/message objects.

anonrig:
Instead of passing a variable to the constructor why don't we use ExpressJs' `app.use()` typing? It seems more understandable and may afford multiple middlewares at the same time. 

What do you think @mikehmorrissey ?

benbrown:
Or both!

mikehmorrissey:
@DEGoodmanWilson @anonrig @benbrown @RafaelCosman  thanks for the feedback! I definitely agree that it needs some naming work to allow for middleware in other parts of the event flow. Does anyone have a suggestion for this particular instance? I'm thinking ``preprocessor`` instead of ``middleware``. 

I think we can easily do both the constructor and the expressJS style. how i currently have it allows for setting middleware both ways. 

```js
const BotController = Botkit.slackbot({
  preprocessor: [normalize, unslang]
})

BotController.preprocessor.use(simplify)
```

anonrig:
The naming `middleware` is more fit to this situation. I really liked this naming style.

```
BotController.use(simplify);
```

or

```
BotController.use(function(data) {} );
```

benbrown:
@anonrig with the .use syntax, how do we tell Botkit at which point in the process the function should be fired?

anonrig:
The function shouldn't be fired actually. We can use event driven programming and have an enum of events in the botkit code. With each specific function, we can fire that event with a compatible callback mechanism where a preprocessed function will have access to.

What do you guys think about it?

RafaelCosman:
I'll leave this one to the experts. @benbrown, @mikehmorrissey?

mikehmorrissey:
@anonrig not sure exactly what you are suggesting. I think my current implementation is pretty simple to use and understand. 

@benbrown @RafaelCosman I just saw an interesting module come out that would have fit perfectly as preprocessing middleware https://blog.beepboophq.com/natural-language-slackbots-with-wit-ai-and-beep-boop-c007cf303c04#.v1egcp1zk

benbrown:
Still thinking about this! I plan on adding in some middleware hooks soon, but I am still researching the best way to do so.

RafaelCosman:
@benbrown what are you thinking?

colestrode:
I'd say if you're going to call them "middleware", which seems appropriate given @mikehmorrissey's first example, then I think will make the most sense to people using Express to do what @anonrig is suggesting. I would have messages pass through each middleware in the order it was registered. 

Whatever syntax you choose, I would suggest picking only one. Having two ways to support the same thing increases your complexity, docs, tests, etc.

dcsan:
is this pattern also going to work for adapting output to various clients? eg Slack may need a message packaged in a different way than hipchat and that can be delegated to the right middleware?

I guess hubot uses the "adapter" pattern for this, but it wasn't flexible enough to support anything other than lowest common denominator see this issue:
https://github.com/github/hubot/issues/1036#issuecomment-178240267

mikehmorrissey:
@benbrown feel free to close this. seems like you have a better approach on the middleware branch!

PR #95, opened 2016-01-21
Adding morgan

ChrisCinelli:
I was trying to make https://github.com/howdyai/botkit/blob/master/examples/slackbutton_bot.js works as an app. The `oauth` was not working. It took a while to find out what was going on. In general it is not a bad idea to be able to see what hit the express endpoints. morgan helps to log that.

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -1,6 +1,7 @@
 var Botkit = require(__dirname + '/CoreBot.js');
 var request = require('request');
 var express = require('express');
+var morgan = require('morgan');
 var bodyParser = require('body-parser');
 
 function Slackbot(configuration) {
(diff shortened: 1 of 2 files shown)


anonrig:
For logging purposes please follow the README section `Logging`

ChrisCinelli:
I think you refer to https://github.com/howdyai/botkit#writing-your-own-logging-module
It seems that that is how to use different logger instead of `console.log`, BTW, I realize that you can only use it for the `botkit.log` but not for `botkit.debug`.

But anyway, `morgan` is actually adding **extra** information that you currently botkit does not log. In particular, morgan, will log when something hit an express end point like (`/login`, `/oauth`, etc).

anonrig:
Thank you for your response.

I am not quite sure if this feature should be optional or should exist inside the Botkit itself.

What do you guys think? @benbrown @RafaelCosman

RafaelCosman:
@ChrisCinelli, thanks for submitting this pull request! @anonrig, this sounds pretty useful to me. I'm inclined to include it.

ChrisCinelli:
You can change where it logs thought the `stream` option. See https://github.com/expressjs/morgan#single-file for an example. You can create a stream and output the output  to a handler that call `botkit.log` or `botkit.debug`. I would have done it, but it is not clear which of the 2 functions should be used (See also #96)

PR #98, opened 2016-01-22
Add pattern match to message object in convos

mgontav:
Adds the pattern matches to the `message` object when handling responses in conversations.

This information is passed along with the message much like it is done with pattern matching in `hears`, saving one additional call to `message.text.match()` when we want to know what exactly were the matching groups (when using regexes).


RafaelCosman:
@mgontav thanks for making this change! We'll test this and then get back to you.

benbrown:
Thanks! Handy.

PR #101, opened 2016-01-24
Make the code more DRY and more flexible using EventEmitter2

ChrisCinelli:
I also add also a way to output a log of the `.trigger` actions (ex: `DEBUG=*,-botkit:events:trigger:tick node bot.js` or `DEBUG=botkit:events:trigger:presence_change node bot.js`).

The code that was doing event management was the same in 3 places (copy and pasted)
It also had a mechanism that prevent the unprocessed listeners to execute if a listener return false. This mechanisms is error prone since it depends on the order in which the event listeners (`.on`) are registered. It seemed that the mechanism was not used beside optimization.

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -8,9 +8,10 @@ var simple_storage = require(__dirname + '/storage/simple_storage.js');
 var ConsoleLogger = require(__dirname + '/console_logger.js');
 var LogLevels = ConsoleLogger.LogLevels;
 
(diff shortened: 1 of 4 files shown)


anonrig:
Very nice @ChrisCinelli 

Thank you for your pull request. Can you write your comments using `JSDoc` syntax?

Additionally, very well written tests. Good job!

ChrisCinelli:
Done

anonrig:
Perfect!

PR #102, opened 2016-01-24
Fix hear

ChrisCinelli:
More performant `hears` function.
It only listen to one event for each hear.
Precompile regEx.
Add the option to have a match function instead of a string
With debug on, we show the rule that matched.

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -610,31 +610,52 @@ function Botkit(configuration) {
     };
 
     botkit.hears = function(keywords, events, cb) {
-        if (typeof(keywords) == 'string') {
-            keywords = [keywords];
-        }
-        if (typeof(events) == 'string') {
-            events = events.split(/\,/g);
-        }
+        var matchFn;
+        var theRule;
 
-        var match;
-        for (var k = 0; k < keywords.length; k++) {
-            var keyword = keywords[k];
-            for (var e = 0; e < events.length; e++) {
-                (function(keyword) {
-                    botkit.on(events[e], function(bot, message) {
-                        if (message.text) {
-                            if (match = message.text.match(new RegExp(keyword, 'i'))) {
-                                botkit.debug('I HEARD ', keyword);
-                                message.match = match;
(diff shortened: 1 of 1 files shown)


anonrig:
Please follow the JSDOC for documentation.

lib/CoreBot.js
```diff
@@ -610,31 +610,48 @@ function Botkit(configuration) {
...
-        if (typeof(keywords) == 'string') {
-            keywords = [keywords];
-        }
-        if (typeof(events) == 'string') {
-            events = events.split(/\,/g);
+        var matchFn;
+        var theRule;
+
+        // Compile
+        if (typeof(keywords) == 'function') {
```
anonrig: I believe this should be `===`

lib/CoreBot.js
```diff
@@ -610,31 +610,48 @@ function Botkit(configuration) {
...
+            matchFn = keywords;
+            // Store first line of the fuction
+            theRule = (function strText(functionText){return functionText.substr(0, functionText.indexOf("\n"));})(keywords)
```
anonrig: There is a 120 length limit per line. Will appreciate if you don't make it a single line.

anonrig:
Test for this function is missing. Please include your tests.

lib/CoreBot.js
ChrisCinelli: both `==` and `===` works. It is a string comparison and  `==` was in the previous code.

lib/CoreBot.js
ChrisCinelli: Fixed

ChrisCinelli:
No function in `CoreBot.js` is tested. If you guys set up tests, I will enhance the test of `hears` to cover the case for `function` as `keywords`.

benbrown:
I have done what is probably a pretty similar refactor of the way hears works, in addition to implementing a new middleware system which could be used to extend hears to receive functions. 

Take a look:
https://github.com/howdyai/botkit/tree/experimental

PR #105, opened 2016-01-26
Fixed linting errors

colestrode:
I noticed that I had several linting errors locally. I ran `jscs ./ --fix` to fix most of them. I had to add a new linting rule for JSDoc to allow the `@callback` annotation. The alternative would be to remove that annotation from lib/events.js.

I also added mocha as a dev dependency so that tests could be run.


anonrig:
Pull request has conflicts. Please reopen a pull request.

PR #117, opened 2016-02-09
Add handling of undefined message object text in conversation.tick

shotmonkey:
When using conversation.say with a message object that had no text but did have an attachment, the undefined text property was still passed to mustache for rendering, throwing a TypeError.

This change adds a simple check that the text exists before attempting to render it, otherwise setting message.text to an empty string (which does not appear in slack output at all - not even a blank line)


benbrown:
Thank you!

PR #118, opened 2016-02-09
Add file header to slackbutton_bot.js example

tristansokol:
Helps users by including the description of the example, as well as the basic instructions to use the example. Namely the `clientId=<my client id> clientSecret=<my client secret> port=3000 node bot.js` command.


benbrown:
Thanks!

Issue #111, opened 2016-01-30
bot.api.chat.postMessage requires JSON.stringify(attachment)


Studnicky:
Just for API Consistency - 

When posting attachments, the bot.say & bot.reply commands do not require a stringified attachment, while bot.api.chat.postMessage does.

paulwehner:
I see that a fix has been put in for bot.api.chat.postMessage. Is there going to be this same issue of needing to stringify the attachment with bot.api.chat.update?

Also is there a reason the readme.md doesn't including documentation related to bot.api.chat.postMessage or bot.api.chat.update? I would like to update a message later using update() as I would be able to do it based on a timestamp without needing the response_url as documented here: https://api.slack.com/docs/message-buttons#use_chat.update_to_modify_the_original_message_instead. Would using bot.api.chat.postMessage and bot.api.chat.update be the best approach for this still?

PR #121, opened 2016-02-15
Documentation for messages with attachements and initiating private conversation

mvaragnat:
This follows several discussions on the chat and discovery of subtle differences between how to pass formatted messages with attachements as input to bot.reply, convo.say and chat.postMessage


mvaragnat:
#111 is an open issue on chat.postMessage. This PR shows an example of what to do, until #111 is solved

benbrown:
Thanks for this!

PR #122, opened 2016-02-15
Fix problem with json parameter for middleware call

xVir:
Fix lack of the `cb` parameter on `bot.middleware.post_api` call.


benbrown:
Thanks, @xVir!

xVir:
:+1:

PR #126, opened 2016-02-17
Post JSON to response URL, not form encoded

wilg:
The slash command `response_url` takes a JSON body, not a form-encoded one. Form encoding will URL encode special characters in the message you're trying to send.

### DIFF
--- a/lib/Slackbot_worker.js
+++ b/lib/Slackbot_worker.js
@@ -280,7 +280,12 @@ module.exports = function(botkit, config) {
             msg.channel = src.channel;
 
             msg.response_type = 'in_channel';
-            request.post(src.response_url, function(err, resp, body) {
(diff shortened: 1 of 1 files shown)


lib/Slackbot_worker.js
```diff
@@ -330,7 +335,13 @@ module.exports = function(botkit, config) {
             msg.channel = src.channel;
 
             msg.response_type = 'ephemeral';
-            request.post(src.response_url, function(err, resp, body) {
+
+            var requestOptions = {
+              uri: src.response_url,
```
anonrig: Did you run a `npm test`? Indentation is `4`.

benbrown:
Thanks @wilg.

@anonrig I'll fix that.

PR #131, opened 2016-02-22
Add mocha as dev dependency

### LINKED ISSUE #130: package.json depends on a global install of mocha
`mocha` should be installed as a `devDependency` to remove global dependencies in the same way that the `package.json` depends on `jscs`.

Seikho:
Resolves #130


anonrig:
Thank you!

PR #137, opened 2016-03-01
use bot returned by startRTM

jmjavier:
If you inspect _bots object the old code, it ends up with only a single key. Using the bot returned by startRTM stores all the bots in _bots correctly.


RafaelCosman:
@jmjavier, thanks for working on this. @benbrown, can you take a look at this when you have a chance?

benbrown:
Good catch!

PR #138, opened 2016-03-01
Adds maximum char length for topic and purpose

RafaelCosman:
I've found it useful to have these constants in the slack api file.


benbrown:
@RafaelCosman you've got some syntax problems in this.

PR #144, opened 2016-03-02
Fix bug that causes `postMessage` to silently fail

ianalexander:
Botkit sends an id with each message sent via `postMessage`. This id is reset after botkit is restarted. Because we were including the message id with the same message, slack would discard these messages with `{ok => true}` because it had just seen an identical message with an identical id. This PR removes the id from the message being sent via `postMessage` so that slack will post each message.

Thanks to @benbrown for tipping me off to this!


benbrown:
I'm going to merge #142 instead of this one because we should still send it when using the RTM.

benbrown:
Thanks again @ianalexander

Issue #145, opened 2016-03-03
RTM API does not work with a proxy


wilg:
This is fixed in https://github.com/slackhq/node-slack-client/pull/91 – perhaps botkit should be based on the official node API to prevent issues like this?

A related issue was closed but is caused by this: https://github.com/howdyai/botkit/issues/21

wilg:
closed with #147

Issue #21, opened 2015-12-19
Unable to connect when behind a proxy 


benmatselby:
We are unable to connect when we need to use a proxy

The http_proxy environment variables are set but this doesn't make a difference 

Should this work behind a proxy?

anonrig:
I've inspected slack_web_api.js file. It uses requestjs for HTTP requests and websockets to connect to slacks websocket api. I believe even with a proxy http requests will work. In theory, websockets seem available for proxy use, but I don't know how `ws` library applies in practice.

Do you have any firewall behind your network which inhibits botkit to use websockets?

@benmatselby

RafaelCosman:
@benmatselby, can you answer @anonrig's question if you have a chance?

benmatselby:
Yes. Sorry about the delay. I am not in the office until Monday so cannot do any checks.

benmatselby:
@anonrig @RafaelCosman 

I've spoken to our networks team and they believe this is something on our side, and are making changes to certain policies, so closing, and will re-open with more information should it still not work

jmkhael:
Hello,
I suffer from the same issue behind our work proxy/firewall. My bot works fine from home where there are no proxying/... going on

Behind the proxy/fw the connection fails (times out on port 443) with botkit whereas the official [node-slack-client](https://github.com/slackhq/node-slack-client) works. (The only difference I noticed is that they use https-proxy-agent in their dependencies...)

I'd like to stay on botkit as the abstractions are nicer :)

PR #147, opened 2016-03-05
Add proxy support for web socket connection

petemichel77:
botkit doesn't work behind a proxy.  The initial call start the RTM session works because the request library automatically checks for HTTP Proxy environment variables.  The websocket library does not do the same thing.  This pull request uses the http-proxy-agent library to add proxy support to the websocket call.  This pull request is for issues #21 (which was closed) and #145.


wilg:
:+1: Awesome, thanks!

PR #154, opened 2016-03-15
clone the config to avoid mutation, attach the access_token to the team and use for web API

phated:
I noticed that, even though I was requesting additional scopes for my bot during the OAuth flow, that any calls made to `bot.api.*` were failing with the response:

```
{ ok: false,
  error: 'missing_scope',
  needed: 'channels:write',
  provided: 'identify,bot:basic' }
```

I followed the rabbit hole and found that the bot token was being assigned as the primary token through mutation of the config object in `configureRTM`.  By cloning the config object, the mutation doesn't happen and then adding the original `access_token` as token on the team, it allows me to make API calls and still accept RTM.


lib/Slackbot_worker.js
```diff
@@ -1,17 +1,21 @@
...
     var bot = {
         botkit: botkit,
-        config: config || {},
+        config: clone(config),
         utterances: botkit.utterances,
-        api: slackWebApi(botkit, config || {})
+        api: slackWebApi(botkit, clone(config))
     };
 
+    var botApi = slackWebApi(botkit, bot.config);
```
anonrig: Why define the second slackWebApi instance? line `9` has  a slackWebApi instance and now this line.

lib/Slackbot_worker.js
```diff
@@ -1,17 +1,21 @@
 var Ws = require('ws');
+var clone = require('lodash.clone');
```
anonrig: Can you use `lodash.clone()` function? Instead of adding a library only serves as a `clone` creator.

phated: I don't see lodash in the dependencies. Would you rather have lodash instead of the modular clone method?

lib/Slackbot_worker.js
phated: The bot API needs the bot token and the web api needs the `access_token` with extra scopes. The `bot.config` object has the bot token.

lib/Slackbot_worker.js
anonrig: Yep, because `lodash` addresses to a more broad audience and functionalities.

phated: sounds good.  Updated and rebased.

anonrig:
@benbrown what do you think?

lib/Slackbot_worker.js
```diff
@@ -1,17 +1,21 @@
...
     var bot = {
         botkit: botkit,
-        config: config || {},
+        config: _.clone(config),
         utterances: botkit.utterances,
-        api: slackWebApi(botkit, config || {})
+        api: slackWebApi(botkit, _.clone(config))
     };
 
+    var botApi = slackWebApi(botkit, bot.config);
```
anonrig: I'm not really quite sure why you didn't use `bot.api` but preferred a different variable exactly as the same one and use `botApi`

anonrig: I believe this is a workaround and a proper fix can be made.

phated: One uses the bot's token and the other uses the user's token.  The mutation in the code makes this change much harder if you don't create 2 API instances.

josephfinlayson:
I've experienced this problem too.

benbrown:
I'd love to merge this, but I'm not 100% sure what causes this issue.

@josephfinlayson @phated can you help me understand?

lib/Slackbot_worker.js
```diff
@@ -1,17 +1,21 @@
...
 
 module.exports = function(botkit, config) {
+    config = config || {};
     var bot = {
         botkit: botkit,
-        config: config || {},
+        config: _.clone(config),
         utterances: botkit.utterances,
-        api: slackWebApi(botkit, config || {})
+        api: slackWebApi(botkit, _.clone(config))
```
colestrode: You want to use _.cloneDeep. _.clone is a shallow clone so any objects or arrays in the `config` object will be referenced both by the original object and the cloned one. This could lead to the same modification problem you're seeing now.

benbrown:
We brought in a dependency on a non-underscore clone package which we can use.

I'm still not quite sure what the cause or effect of this is.

rlau1115:
+1, any update on a fix for this?

emhagman:
I also appear to be having this issue. Using RTM and API

jonchurch:
Eric are you on the latest botkit version?
Not sure if that should matter, but curious.

PR #157, opened 2016-03-20
Add bot.destroy() to prevent memory leaks

mbrevoort:
Currently there is no way to properly cleanup a spawned worker. This PR adds `bot.destroy()` to cleanup a worker. `destroy()` was added instead of overloading `bot.closeRTM()` since `spawn` is distinct from `bot.startRTM()` and you may want to close and RTM connection and reconnect without tearing down the worker.

leak

Update: I also added logic to detect stale connections if no `PONG` messages have been received.

### DIFF
--- a/lib/Slackbot_worker.js
+++ b/lib/Slackbot_worker.js
@@ -12,6 +12,8 @@ module.exports = function(botkit, config) {
         api: slackWebApi(botkit, config || {})
     };
 
+    var pingIntervalId = null, lastPong = 0;
+
     /**
      * Set up API to send incoming webhook
      */
@@ -46,9 +48,26 @@ module.exports = function(botkit, config) {
         return bot;
     };
 
-    bot.closeRTM = function() {
-        if (bot.rtm)
+    bot.closeRTM = function(err) {
+        if (bot.rtm) {
+            bot.rtm.removeAllListeners()
             bot.rtm.close();
+        }
+
+        if (pingIntervalId) {
+            clearInterval(pingIntervalId);
+        }
+
+        lastPong = 0
+        botkit.trigger('rtm_close', [bot, err]);
+    };
+
+    /**
+     * Shutdown and cleanup the spawned worker
+     */
+    bot.destroy = function() {
+        bot.closeRTM()
+        botkit.shutdown()
     };
 
     bot.startRTM = function(cb) {
@@ -86,10 +105,22 @@ module.exports = function(botkit, config) {
             bot.rtm = new Ws(res.url, null, {agent: agent});
             bot.msgcount = 1;
 
-            var pingIntervalId = null;
+            bot.rtm.on('pong', function (obj) {
+                lastPong = Date.now()
+                botkit.debug('PONG received');
+            })
+
             bot.rtm.on('open', function() {
 
                 pingIntervalId = setInterval(function() {
+                    if (lastPong && lastPong + 12000 < Date.now()) {
+                        var err = new Error('Stale RTM connection, closing RTM')
+                        botkit.log.error(err);
+                        bot.closeRTM(err)
+                        return
+                    }
+
+                    botkit.debug('PING sent');
                     bot.rtm.ping(null, null, true);
                 }, 5000);
 
@@ -121,6 +152,9 @@ module.exports = function(botkit, config) {
 
             bot.rtm.on('error', function(err) {
                 botkit.log.error('RTM websocket error!', err);
+                if (pingIntervalId) {
+                    clearInterval(pingIntervalId);
+                }
                 botkit.trigger('rtm_close', [bot, err]);
             });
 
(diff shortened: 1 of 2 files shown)


onlymejosh:
Nice! I was going to work on something similar. Is there an easy way of testing this?

mbrevoort:
Ultimately it would be great to have a benchmark suite to test this scenario with simulated socket connections, but for now I tested locally with the code below, connecting and disconnecting over and over and watching the memory usage. And then I tested in a Beep Boop staging environment with 40 team connections going through a similar process.

```javascript
// for taking heap snapshots
require('heapdump')
var bot

function connect () {
  bot =  controller.spawn({
      token: token,
      debug: true
    }).startRTM(function(err,bot,payload) {
      if (err) {
        throw new Error(err);
      }
      setTimeout(disconnect, 10000)
    });
}

function disconnect () {
  bot.destroy()
  bot = null
  setTimeout(connect, 5000)
}

function probe () {
  console.log(process.memoryUsage())
}

setInterval(probe, 5000)
connect()
```

lib/Slackbot_worker.js
```diff
@@ -86,10 +105,22 @@ module.exports = function(botkit, config) {
...
-            var pingIntervalId = null;
+            bot.rtm.on('pong', function (obj) {
+                lastPong = Date.now()
+                botkit.debug('PONG received');
+            })
+
             bot.rtm.on('open', function() {
 
                 pingIntervalId = setInterval(function() {
+                    if (lastPong && lastPong + 12000 < Date.now()) {
```
dossy: Since `lastPong` is initialized to `0` (line 15, line 61), and through my testing it appears that Slack's websocket server does NOT respond to WS "ping" messages with a "pong" ... `lastPong` will always be `0` and this `if` condition will always evaluate to false.

Have you ever actually observed the error inside this `if` conditional occurring?

mbrevoort: Slack's RTM websocket connection didn't support ping/pong in the past; however, it does today. I tested this change quite a bit with up to 40 socket connections at a time and various failure scenarios.

lib/Slackbot_worker.js
```diff
@@ -86,10 +105,22 @@ module.exports = function(botkit, config) {
...
+                        var err = new Error('Stale RTM connection, closing RTM')
+                        botkit.log.error(err);
```
anonrig: Unified error handling and passing to log system is mandatory for better maintenance. Therefore, please update this line like line 154. `botkit.log.error('RTM websocket error!', err);`

readme.md
```diff
@@ -227,7 +227,35 @@ bot.startRTM(function(err,bot,payload) {
   if (err) {
     throw new Error('Could not connect to Slack');
   }
+
+  // close the RTM for the sake of it in 5 seconds
```
anonrig: Please read CONTRIBUTING.md before creating a pull request. JSDoc is required.

anonrig: Please don't add obvious comments.

readme.md
```diff
@@ -227,7 +227,35 @@ bot.startRTM(function(err,bot,payload) {
...
+  token: my_slack_bot_token
+})
+
+bot.startRTM(function(err, bot, payload) {
+  if (err) {
+    throw new Error('Could not connect to Slack');
+  }
 });
+
+// some time later (e.g. 10s) when finished with the RTM connection and worker
```
anonrig: Great example of an obvious comment.

readme.md
mbrevoort: This is in the `readme.md`, you want JSDoc style comments in the README? Also *I DID* read CONTRIBUTING.md before submitting and it doesn't mention JSDoc at all.

readme.md
mbrevoort: This too is in the REAME. I was attempting to make the example I added to the doc more narrative. I can remove it if you really want.

readme.md
anonrig: ```Follow the JavaScript coding style with details from .jscsrc and .editorconfig files and use necessary plugins for your text editor.```

readme.md
colestrode: I find these comments in the readme to be helpful since the goal of the readme is to walk users through what is happening and why.

lib/Slackbot_worker.js
dossy: BTW, you're right -- I was doing it wrong.  `bot.rtm.on()` was the right method - I was peeking inside the `ws` module, and tried doing `bot.rtm._receiver.onpong()` directly, which wasn't working.

Nice work!

alecl: Hmm. If you are interactively debugging your bot and stepping through for a bit you will miss the pong and then get this Stale RTM killing the connection. Any thoughts on whether bypassing the pong if in debug mode is a valid idea?

// http://stackoverflow.com/a/13454643
var debug = typeof v8debug === 'object';

mbrevoort:
Interesting scenario. Yes, I think it's valid; however you'll still
potentially have the problem of the TCP connection going stale. Disabling
the explicit ping/pong would at least alleviate the issue cause by stopping
the world though.

alecl:
Cool. I'll give it a shot this week and see what else happens downstream.
Maybe I can also put in a reconnect if the connection goes stale on its own.

PR #166, opened 2016-03-29
Use the same global pattern matching function inside convo.ask as is used at controller.hears

benbrown:
Enable developer to change pattern matching function at a global level using controller.changeEars()
Strip @mentions from beginning of messages that are part of a conversation
Add additional .ask examples in bot.js


benbrown:
Good job, me!

PR #167, opened 2016-03-31
Adds authentication of incoming Slack requests

sgud:
Adds optional support for authenticating requests from
Slack slash commands.  Validates that the request body
includes a known token.

### DIFF
--- a/lib/middleware/authentication.js
+++ b/lib/middleware/authentication.js
@@ -1,0 +1,41 @@
+/**
+ * Authentication module composed of an Express middleware used to validate
+ * incoming requests from the Slack API for Slash commands and outgoing
+ * webhooks.
+ */
+
+// the list of authorized tokens that can invoke slash comamnds
+var authenticationTokens = [];
+const TOKEN_NOT_FOUND = -1;
+
+function init(tokens) {
+    authenticationTokens = authenticationTokens.concat(tokens);
+    return authenticate;
+}
+
+/**
+ * Express middleware that verifies a Slack token is passed;
+ * if the expected token value is not passed, end with request test
+ * with a 401 HTTP status code.
+ *
+ * Note: Slack is totally wacky in that the auth token is sent in the body
+ * of the request instead of a header value.
+ *
+ * @param {object} req - Express request object
+ * @param {object} res - Express response object
+ * @param {function} next - Express callback
+ */
+function authenticate(req, res, next) {
+    if (!req.body || !req.body.token || authenticationTokens.indexOf(req.body.token) === TOKEN_NOT_FOUND) {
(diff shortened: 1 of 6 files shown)


benbrown:
Good idea! Thanks for the PR.

Should this be on by default perhaps? What would that involve?

sgud:
This can certainly be activated by default. I think it makes sense to activate it by default as you can ensure that all endpoints are only ever reached if the token is valid, thus securing the server more effectively. This is in contrast to the current security mechanism which involves validating the api token in the slash command event itself. 

In order to activate by default, the secureWebhookEndpoints() function needs only to be added as the first line of createWebhookEndpoints() and the parameters of createWebhookEndpoints() modified to include authentication tokens.

sgud:
I've modified the PR such that webhook authentication is now a default feature when creating webhook endpoints. Furthermore, by leaving the secureWebhookEndpoints() function as optional, this shouldn't impact backwards compatibility.

lib/middleware/authentication.js
```diff
@@ -0,0 +1,63 @@
+/**
+ * Authentication module composed of an Express middleware used to validate
+ * incoming requests from the Slack API for Slash commands and outgoing
+ * webhooks.
+ */
+
+// the list of authorized tokens that can invoke slash comamnds
+const TOKEN_NOT_FOUND = -1;
```
anonrig: Can you replace `const` with `var` please? We're trying to support NodeJS 0.12.x also.

anonrig:
What a nice PR! Congrats.

lib/middleware/authentication.js
sgud: Done, I've replaced const with var and fixed the comment in line 7.

sgud:
@benbrown @anonrig:
Just wanted to check and see if there are any other changes you need made to the PR before it can be merged in. Please let me know if anything comes up.

steve-jansen:
FYI if you want to test drive this PR.

I did an npm pack build locally of this PR after rebasing from upstream/master (commit 940c3e0).

Usage:

```
npm install https://github.com/steve-jansen/botkit/releases/download/v0.0.15-request-token-auth/botkit-0.0.15.tgz
```

benbrown:
I'll merge this in ASAP!
Sorry for the delay - just had a kid!

steve-jansen:
@benbrown congrats!  I'm a proud dad of 3 these days... 

The days are long, the years are short. Enjoy them!

sgud:
@benbrown Congratulations on the baby, wish you the best!

benbrown:
This is fabulous! I couldn't do a direct merge of this PR, but all of this code and these new features are now included!

PR #174, opened 2016-04-06
Make identity available to the user of the OAuth endpoint.

sorribas:
When a user gets redirected back to my app from slack I want to redirect them to a URL called `/team/:team-id`, but I don't have access to the team id. This adds the identity to the request object so that I can access it in the callback.


sorribas:
Any comments on this?

freeall:
We also do this in our fork of botkit.

Could this be merged?

Issue #186, opened 2016-04-16
Typo in Read Me


adbierach:
The Read Me begins "Botkit designed"...it should probably read "Botkit is designed"

PR #187, opened 2016-04-16
Fix typo in read me

adbierach:
Changes "Botkit designed" to read "Botkit is designed".
Resolves issue #186


anonrig:
Thank you!

PR #189, opened 2016-04-17
adds support for <@userId|username> syntax for mentions and direct_mentions

colestrode:
For Slack, the RegExp for mentions and direct_mentions does not match if the message uses the valid <@userId|username> syntax. This PR updates the regex so that botkit will emit mention and direct_mention events if the username is included.

Note: usernames can only be in the set [a-z0-9] so there won't be any special characters that could cause problems constructing a regex including usernames. See https://get.slack.help/hc/en-us/articles/216360827-Changing-your-username

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -416,6 +416,9 @@ function Slackbot(configuration) {
 
         slack_botkit.log('** Setting up custom handlers for processing Slack messages');
         slack_botkit.on('message_received', function(bot, message) {
+            var mentionSyntax = '<@' + bot.identity.id + '(\\|' + bot.identity.name + ')?>';
+            var mention = new RegExp(mentionSyntax, 'i');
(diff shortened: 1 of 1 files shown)


crunchywelch:
related, other option for a way to do this in #210 

fine with either way, just want to be able to get events on nickname mentions as well

lib/SlackBot.js
```diff
@@ -416,6 +416,9 @@ function Slackbot(configuration) {
 
         slack_botkit.log('** Setting up custom handlers for processing Slack messages');
         slack_botkit.on('message_received', function(bot, message) {
+            var mentionSyntax = '<@' + bot.identity.id + '(\\|' + bot.identity.name + ')?>';
```
colestrode: Looks like usernames can contain ".", we'll need to escape those

PR #190, opened 2016-04-17
Fix Typo: "handled using by" -> "handled by"

ariannaisrael:
##### Fix Typo:  

"These message events can be handled <strike>using</strike> by attaching an event handler to the main controller object."  

In **readme.md > Developing with Botkit > Receiving Messages**


anonrig:
Thank you!

PR #195, opened 2016-04-19
Save users auth and team even if it is not new

Dinoshauer:
We just ran into a "thing" (it's not really a bug I suppose), where the users auth token wouldn't be saved if the user already existed in the storage.

The `auth_token`, `scopes`, and others can safely be updated when the user is updated. So this PR takes care of that.

In our case what happens was that: bot saves all the users in the storage upon connecting to the `rtm` to save on api calls to slack and minimize latency, so if a user removes our app (is there an event for that even?) and we haven't removed the slack data before another user of the same team adds the app again, the access_token and scope list will be missing from the user who authenticated the app last.

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -315,16 +315,13 @@ function Slackbot(configuration) {
 
                             slack_botkit.findTeamById(identity.team_id, function(err, team) {
 
-                                var isnew = false;
-                                if (!team) {
-                                    isnew = true;
-                                    team = {
-                                        id: identity.team_id,
-                                        createdBy: identity.user_id,
-                                        url: identity.url,
-                                        name: identity.team,
-                                    };
-                                }
+                                var isnew = !!team;
+                                team = team || {};
+
+                                team.id = identity.team_id;
+                                team.createdBy = identity.user_id;
+                                team.url = identity.url;
+                                team.name = identity.team;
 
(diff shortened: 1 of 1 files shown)


anonrig:
Hello @Dinoshauer 

Thank you for your merge request.

The correct functionality to handle these kinds of situations is I believe mongoose's way of thinking. Adding a `{ upsert: true }` option to `save` method, seems to give more functionality rather than just removing the ability to have static/constant user data.

Dinoshauer:
I'm not sure I follow, if you added an `{ upsert: true }` option to the `save` method the `access_token` etc would never get updated regardless as the object is not constructed/updated if the entity already exists

freeall:
@anonrig To me it makes a lot of sense that when you call a `get()` then you actually get the user data, and not an old version of that user. Right now, if e.g. the `access_token` has changed then that will not be seen in the data returned.

The only way an options map with `{upsert: true}` would make sense is for the `slack_botkit.findTeamById(..)` and `slack_botkit.storage.users.get(..)` methods. But even then I'm not sure it's a good way.

rokoroku:
I suffered from same issue. I also using modified code like Dinoshauer's idea. There will be a lot of cases that you can easily imagine if we don't update registered user's information.

benbrown:
This has been taken care of in a slightly different way.

scopes and token should always be saved.

PR #199, opened 2016-04-20
Move "MESSAGE HOOK" message into debug output.

shindakun:
no issue
- Move "GOT A MESSAGE HOOK" message into debug output instead of console.log.


anonrig:
Nice catch!

PR #203, opened 2016-04-25
Add remove method for redis_storage.js

andrewholsted:
I needed the ability to remove teams/users from Redis. I added the method to the example file here, but didn't add it for the other types of storage (simple and firebase). I can explore adding it to the other methods to make this pull request complete if there is enough interest. I wasn't sure why it wasn't included in the first case so I could be missing some context. Open to thoughts/suggestions/improvements.

### DIFF
--- a/lib/storage/redis_storage.js
+++ b/lib/storage/redis_storage.js
@@ -34,6 +34,9 @@ module.exports = function(config) {
                         return cb(new Error('The given object must have an id property'), {});
                     client.hset(config.namespace + ':' + hash, object.id, JSON.stringify(object), cb);
                 },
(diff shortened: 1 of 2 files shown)


anonrig:
You should add this function to botkit-storage-redis. Not here.

https://github.com/howdyai/botkit-storage-redis

colestrode:
Seems to me this is part of a larger conversation about supporting delete. I would think botkit itself would need to be updated to support delete, right? At least the storage_test files. And if it were included for Redis, shouldn't it be included for the other storage libs?

Personally, I really like the idea of having a delete method for the storage interfaces, it's always seemed missing to me.

PR #210, opened 2016-05-03
allow for matching on any nickname mention

crunchywelch:
sometimes we like to alert in channel with just the nick, without a proper `@nickname` mention, this allows for a broader match for any mention of the nick such as `nickname: message` this patch adds a `broad_mention` message event type which catches those instances


anonrig:
Hey @crunchywelch 

Thank you for your pull request. Honestly, it's really great the ability to send a broad_mention event upon a message that includes a username, but I am not really quite sure if this functionality should be added to the library itself or should be added to an example project itself.

What about we wait for other responses/thoughts in this manner? :)

benbrown:
Let's use the version of this functionality in this PR:

https://github.com/howdyai/botkit/pull/189

Issue #136, opened 2016-02-29
Files from public aren't served as expected


ToddHoff:
I'm not sure if this is a bug or just my misunderstanding. I thought files would be served out of /public in my bot directory because of in SlackBot.js :

slack_botkit.webserver.use(express.static(__dirname + '/public'));

This doesn't work because __dirname resolves to "bot/node_modules/botkit/lib", not my bot directory.

So I added to my code:
webserver.use(require('express').static(__dirname + '/public'));

Now the webserver is finding static files in my public directory.

anonrig:
Nice catch! Can you send a pull request about this? @ToddHoff

ToddHoff:
I'm not exactly sure of the proper fix. I would guess use process.cwd()?

m4olivei:
This was fixed for Slack in #214, but Facebook got left behind :(.  Here is a pul request to remedy that: https://github.com/howdyai/botkit/pull/252

peterswimm:
seemingly resolved, please reopen if this is not the case.

PR #214, opened 2016-05-08
Webserver static dir can be configured through configuration

fabito:
The idea is to make it possible to configure the webserver static dir.
The variable `__dirname` points to the directory where the SlackBot.js file resides which is usually inside `node_modules`.

```
var controller = Botkit.slackbot({
      webserver: {
        static_dir: __dirname + '/public'
      }
    });
```
An alternative (which actually seems better) is to add a new parameter to the `setupWebserver` function.

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -171,7 +171,7 @@ function Slackbot(configuration) {
         slack_botkit.webserver = express();
         slack_botkit.webserver.use(bodyParser.json());
         slack_botkit.webserver.use(bodyParser.urlencoded({ extended: true }));
-        slack_botkit.webserver.use(express.static(__dirname + '/public'));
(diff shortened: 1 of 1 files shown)


fabito:
Seems like it solves #136

anonrig:
Thanks @fabito, can you also update the docs?

fabito:
Sure.

First I will move the configuration parameter to the `setupWebserver` function.
I was thinking on adding a `config` param which would hold the existing `port` plus the `static_dir`.
Something like this:

```javascript
var config = {
    port: 8080,
    static_dir: __dirname + '/public'
};
controller.setupWebServer(config, cp);
```
If backward compatibility is an issue, we could check whether config is an object or a number and handle both cases.

What do you think ?

anonrig:
Seems good to me. @benbrown what do you think?

Issue #219, opened 2016-05-10
[Facebook] Conversation crashes when big emoji or like is sent


crummy:
1. Start a conversation with a bot (i.e. in a `bot.startConversation()` section, when prompted by `convo.say()`)
2. Click "Like" to send a big thumbs up, or click the :) button and send a large facebook emoji thing

Behaviour: Bot loops TypeError and crashes

Here's my listener:

```
controller.hears("", 'message_received', function(bot, message) {
  bot.startConversation(message, function(err, convo) {
    convo.ask("send me a facebook emoji image to crash me. they are under the `:)` on the web interface. or just click the like button", function(response, convo) {
      convo.say("guess you didn't send me an emoji. instead I got: " + response.text);
      convo.next();
    });
  });
});
```

The crashlog:
```
GOT A MESSAGE HOOK
TypeError: Cannot read property 'trim' of undefined
    at Conversation.capture (C:\Users\Malcolm\Dropbox\code\janbot\node_modules\botkit\lib\CoreBot.js:64:42)
    at Conversation.handle (C:\Users\Malcolm\Dropbox\code\janbot\node_modules\botkit\lib\CoreBot.js:96:22)
    at C:\Users\Malcolm\Dropbox\code\janbot\node_modules\botkit\lib\CoreBot.js:871:31
    at Object.bot.findConversation (C:\Users\Malcolm\Dropbox\code\janbot\node_modules\botkit\lib\Facebook.js:97:25)
    at C:\Users\Malcolm\Dropbox\code\janbot\node_modules\botkit\lib\CoreBot.js:869:21
    at next (C:\Users\Malcolm\Dropbox\code\janbot\node_modules\ware\lib\index.js:82:27)
    at Ware.run (C:\Users\Malcolm\Dropbox\code\janbot\node_modules\ware\lib\index.js:88:3)
    at Object.Botkit.botkit.receiveMessage (C:\Users\Malcolm\Dropbox\code\janbot\node_modules\botkit\lib\CoreBot.js:864:35)
    at C:\Users\Malcolm\Dropbox\code\janbot\node_modules\botkit\lib\Facebook.js:136:45
    at Layer.handle [as handle_request] (C:\Users\Malcolm\Dropbox\code\janbot\node_modules\express\lib\router\layer.js:95:5)
```

Botkit `0.1.1`
Windows 10, ngrok
Node `v5.4.0`

benbrown:
This code has been merged in! Thanks for helping to fix this.

PR #220, opened 2016-05-10
Fix crash upon receiving Facebook emoji or like in conversation

crummy:
See https://github.com/howdyai/botkit/issues/219

I'm not very familiar with JS so there's likely to be a better way to do this, but this hack prevents the crash detailed in the issue above.

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -859,7 +859,7 @@ function Botkit(configuration) {
             // set up a once a second tick to process messages
             botkit.tickInterval = setInterval(function() {
                 botkit.tick();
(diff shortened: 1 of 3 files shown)


benbrown:
Hey @crummy  @tkornblit I have added the key features from these pull requests. Since there's some _other stuff_ in here, I'm not going to officially merge this PR. BUT, master now includes the fix for allowing attachment in convos, as well as the ability to hear a postback as a response to a convo.ask! Cool stuff good job.

crummy:
Thanks, I didn't realise adding to my branch after a pull request would add to the pull request itself.

PR #229, opened 2016-05-18
Add domentation for startPrivateConversation()

andrewholsted:
Also make it known that you must pass a text property to all message objects, even if it's only empty string


benbrown:
Thanks!!

PR #242, opened 2016-05-28
Remove unused variable in CoreBot.js

danielsbird:
Remove an unused variable from the `handle` method in CoreBot.js


anonrig:
Nice catch!

Issue #159, opened 2016-03-22
Setup Webserver on RedHat OpenShift


airlikidh2:
Hello,
I've tried uploading my botkit project on RedShift but it was not working.
I had the following error:
**Application 'appname' failed to start (port 8080 not available)**

The problem is that the express webserver needs an IP on RedShift to get started.

What I've done is I added the IP in the SlackBot.js at line 160 to get the IP from the environment.

```
......
        slack_botkit.webserver = express();

        slack_botkit.webserver.set('ip', process.env.OPENSHIFT_NODEJS_IP || "127.0.0.1");

        slack_botkit.webserver.use(bodyParser.json());
        slack_botkit.webserver.use(bodyParser.urlencoded({ extended: true }));
        slack_botkit.webserver.use(express.static(__dirname + '/public'));

        var server = slack_botkit.webserver.listen(
           slack_botkit.config.port,slack_botkit.webserver.set('ip'), ......

```

I'm not sure if it is the case on other hosting services but this might be worth integrating the **ip** parameter when calling `controller.setupWebserver(port,function(err,webserver)`
Works like a charm otherwise.

More info here: http://stackoverflow.com/questions/31358992/application-appname-failed-to-start-port-8080-not-available-on-open-shift-no

martip-sainsburys:
+1

Same thing happened to me, I had to create my own express server.

rithvikvibhu:
@airlikidh2 Did you edit SlackBot.js after deploying (and then editing with nano), or make changes before pushing?

rithvikvibhu:
@airlikidh2 Did you edit SlackBot.js after deploying (and then editing with nano), or make changes before pushing?

airlikidh2:
I edited before pushing/deploying

benbrown:
We've added a `hostname` parameter to the Botkit constructor which should solve this issue.

var controller = Botkit.slackbot({hostname: '123.123.123.123'});

Let me know if there are any issues with this!

airlikidh2:
Thanks Ben, I'll try it out :)

2016-11-07 15:02 GMT-08:00 Ben Brown <notifications@github.com>:

PR #249, opened 2016-06-01
Add support for ip on webserver

jlsjonas:


benbrown:
I think this value should go into the configuration, not as a separate parameter.

jlsjonas:
@benbrown you're right, changes reflected in last commit

lib/SlackBot.js
```diff
@@ -209,6 +209,7 @@ function Slackbot(configuration) {
             static_dir = slack_botkit.config.webserver.static_dir;
 
         slack_botkit.config.port = port;
+        slack_botkit.config.ip = slack_botkit.config.ip || 127.0.0.1;
```
anonrig: `127.0.0.1` should be a string. What do you think @jlsjonas ?

LoneRifle:
Just to help out with paperwork, this issue will resolve #159

anonrig:
@LoneRifle We're waiting a response from @jlsjonas due to a bug on the code. Will soon asap when its resolved.

LoneRifle:
I notice this only takes care of slack and not fb messenger. If need be, I wouldn't mind submitting a PR that encompasses the above changes (and reflects @jlsjonas authorship of course) with the suggested bugfix as well as equivalent ones for FB. fwiw, i'm more interested in the latter. 

I'll come back to this thread in about a month and see where we should go.

benbrown:
I have added support for the `hostname` configuration option to _all_ of the botkit connectors.

Note it is  `hostname` not `ip` since this is what the parameter is named in the Express documentation.

PR #254, opened 2016-06-02
Added more data to slack bot debug messages

ourcach:
Added BOT ID to slackbot debug messages so that, when running multiple bots, you know what's happening with each of them.

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -670,7 +670,7 @@ function Slackbot(configuration) {
                 return false;
             }
 
-            slack_botkit.debug('DEFAULT SLACK MSG RECEIVED RESPONDER');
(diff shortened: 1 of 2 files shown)


benbrown:
This is a good idea, BUT, sometimes bot.identity won't exist, or bot.identity.name won't be set.

Can you include some defensive code for that?

alecl:
@benbrown when does bot.identity/bot.identity.name not exist? I ran into a situation #683 where I had to manually create the identity for a Slack bot worker specially spawned on startup but it was very oddy that I couldn't get it to happen reliably. Sometimes I had the identity and then when I went back to repeat later could not. I ended up putting in code to check for it and create if not but would love to understand it better.

Issue #222, opened 2016-05-13
Newbie Question: Multiple accounts


paulogr:
Is there a way to run multiple accounts in a single server?

The way that was implemented and documented looks like just one facebook account per instance.

Is that correct?

benbrown:
@paulogr as is, the code is set up to handle one Facebook bot. This is correct.

We _should_ support multiple accounts, however. This is something that should be addressed.  It would require the configuration options be passed into the controller.spawn function instead of the controller.facebookbot initialization.

m4olivei:
Would love to see this in botkit!  In my case I'm building a bot for a media outlet with many brands, all with their own FB Page, that all want very similar functionality.

I'm going to dig in a bit and see if I can help here.  Although my node foo is weak.

anonrig:
For the sake of simplicity and ease in resource management, each botkit instance should run over at different processes. Therefore, a manager class that can basically observe the two botkit instances can be done, without the support of a multiple account feature.

m4olivei:
@anonrig not sure I follow, what do you mean by:

> a manager class that can basically observe the two botkit instances can be done

Forgive my ignorance.  Do you maybe have some links/examples that you could point me at explaining such things?

anonrig:
Don't worry @m4olivei 

I was thinking a solution to instantiate two different botkit instances using a super class which has the observer design pattern. (An example: http://www.tutorialspoint.com/design_pattern/observer_pattern.htm)

m4olivei:
@anonrig I see so your still in favour of making this possible, you're just suggesting a certain implementation path?  Is the solution your suggesting something that would ship with Botkit out of the box, or is this something that is the responsibility of me, as a user of the botkit package?

anonrig:
I don't think that Botkit should allow multiple accounts implementation. I believe that the main purpose of Botkit will change if this feature is implemented. If you have any further questions about how to implement such Observer pattern for Botkit instances, feel free to send me a message from dev4slack.slack.com

PR #259, opened 2016-06-03
Support handling Multiple Facebook pages

m4olivei:
This is a crack at supporting multiple Facebook pages. The idea here is that the page_id is passed along in the message object and then used to lookup the corresponding access_token to use when sending messages in reply to Facebook.  You could also use hears middleware to check against the page_id in the incoming message, and handle or ignore the message based on it.  In that way you could essentially run two different bots from the same code base and node process.

Just throwing this out there to see what people think.  I've also updated the Facebook README to explain how to configure this.  But basically, it adds a `access_tokens` config setting passed to the `Botkit.faceboookbot` function, which is an array of objects so you can specify which page_ids you want to listen to and the corresponding access_token.

There was some initial discussion about this feature in https://github.com/howdyai/botkit/issues/222. I know it was turned down, but I wanted to give it a shot anyway. So far, it's working well for my purposes. More testing to be done though.

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -242,6 +242,9 @@ function Botkit(configuration) {
             } else {
                 message.channel = this.source_message.channel;
             }
+            if (this.source_message.page_id) {
+                message.page_id = this.source_message.page_id;
+            }
 
             if (!this.topics[topic]) {
                 this.topics[topic] = [];
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -242,9 +242,6 @@ function Botkit(configuration) {
             } else {
                 message.channel = this.source_message.channel;
             }
-            if (this.source_message.page_id) {
-                message.page_id = this.source_message.page_id;
-            }
 
             if (!this.topics[topic]) {
                 this.topics[topic] = [];
--- a/lib/Facebook.js
+++ b/lib/Facebook.js
@@ -9,6 +9,13 @@ function Facebookbot(configuration) {
     if (!configuration.access_token && !configuration.access_tokens) {
         throw new Error('One of access_token or access_tokens must be set.');
     }
+    // Backwards compatibility with the access_token configration setting.
+    if (!configuration.access_tokens) {
+        configuration.access_tokens = [{
+            page_id: 'default',
+            access_token: configuration.access_token,
+        }];
+    }
 
     // Create a core botkit bot
     var facebook_botkit = Botkit(configuration || {});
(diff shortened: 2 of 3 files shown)


benbrown:
Thank you for the pull request! We need to carefully review this!

paulogr:
Hi ! Good to see this topic being discussed.

Give my two cents about it, I though that this could be more open.

1) The **subscribed_apps** endpoint could not be called from the **setupWebserver** method. Or could be another public method to the users subscribe their own new pages.

2) The **access_token** is used to send a message back to messenger, may be the **access_token** could be easily passed as a parameters on **reply** and say **methods**. So the implementations to get the access_token is user responsibility.

Well I just forked the project and will try to make a PR for this.

Thanks!

jonchurch:
This assumes that you have all the access_tokens to pass in when first running the bot, correct? 

I have been thinking a lot about supporting multiple facebook pages recently, and I like the idea of bundling the functionality to use facebook login to authenticate and subscribe a page to the bot. 
Like how we setup oauth endpoints for slack apps. Facebook Apps can subscribe to many different facebook pages at their Messenger Webhook, so in theory you could have one bot app that was setup across N teams. You could load Botkit Studio scripts based on the page_id the message originates from, so each bot would have unique content and interactions, managed in Botkit Studio. Not aware of anyone offering their bot as a service on Facebook, _yet_ 😉 


I was matching the page_id received at the webhook with the corresponding bot identity and access token in storage, and passing a bot with that identity into `receiveMessage()` so there was page context available everywhere the bot was.

peterswimm:
This will now require some revisions to bring it up to current master before merging.

acrollet:
Hi folks, I've submitted an updated PR - it mostly follows @m4olivei's approach, with the following changes (similar to what @paulogr suggests):

1. a webhook subscription is not automatically added for all pages, but instead a method is made available. (For backwards compatibility, the subscription is automatically added if a single access_token is passed)
2. instead of a hard-coded page_id/access_token mapping, a callback parameter is accepted. This way new pages can be linked without restarting the app.

Hoping this can get in!

goodpic:
Great to see @acrollet is working on the latest codebase on this issue. Thank you for the effort!
I am very keen to use this functionality. Is this expected to be merged soon? Or needs some more works and time?

jsalwen:
any update on this or other approaches to support multiple pages? is somebody willing to merge this if all the conflicts are addressed?

deathg0d:
Any update to resolving this?

maxmeinberg:
I'm also interested in this change, any update?

danieledp:
Hey guys I had the same problem and found a possible solution. Have a look here https://github.com/danieledp/botkit-multi-fb-pages

MagSag:
Does botkit 0.7.0 support  handling Multiple Facebook pages now ?

benbrown:
No. We are not planning on supporting multi-tenancy in the 0.7 branch.

However, the plan is to support this in the next major revision. See #1576.

freta:
Hey @benbrown when will this major revision be out ?

PR #260, opened 2016-06-03
Replace consoleLogger and log solutions with debug

anonrig:
The main purpose of this pull request is to change the logging mechanism from scratch.

`DEBUG=* npm start` will log all of the components to the terminal with different colors.

Example:

screen shot 2016-06-03 at 9 53 01 pm

PS: Have to bump version in order to accept this PR.

### DIFF
--- a/lib/Slack_web_api.js
+++ b/lib/Slack_web_api.js
@@ -5,7 +5,8 @@ var log = require('debug')('slack:api:log');
 
 module.exports = function(bot, config) {
     debug.log = function() {
-        bot.logger.log.apply(bot.logger, ['info'].concat(arguments));
+        if (bot.logger && bot.logger.log)
+            bot.logger.log.apply(bot.logger, ['info'].concat(arguments));
     };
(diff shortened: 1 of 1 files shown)


rafaelamorim:
+1 on this!
you rock man!

derrickpelletier:
Would like to see this instead of the logging built in, but wouldn't it make more sense to have the debug tokens be prepended with something unique to botkit? `botkit:core:bot:debug`?

Just so when using the lib with other `debug`able libs you have the ability to switch. Express for example is prepended with `express:*`, socket.io, nodemon, etc., all do the same.

... or is this PR abandoned in favour of something else that came afterward.

anonrig:
You're right @derrickpelletier 👍 

This is an abandoned PR because none of the other contributors seemed interested on the topic of changing something that's been used in a very long time.

benbrown:
I'd love to re-address this later. The logging is very inconsistent and not great!

PR #264, opened 2016-06-08
Fix typo for Facebook unexpected messages

krismuniz:
Fix typo in [line 184](https://github.com/krismuniz/botkit/blob/master/lib/Facebook.js#L184) of `botkit/lib/Facebook.js` when logging unexpected messages.

It says:
```
botkit.log('Got an unexpected message from Facebook: ', facebook_message);
```

It should say:
```
facebook_botkit.log('Got an unexpected message from Facebook: ', facebook_message);
```

Apparently Facebook is now sending a new type of event/message (presumably read receipt events). This crashes the application and shows an error: `ReferenceError: 'botkit' is not defined`

**This should be solved as soon as possible.**

### DIFF
--- a/lib/Facebook.js
+++ b/lib/Facebook.js
@@ -181,7 +181,7 @@ function Facebookbot(configuration) {
                             facebook_botkit.trigger('message_delivered', [bot, message]);
 
                         } else {
-                            botkit.log('Got an unexpected message from Facebook: ', facebook_message);
+                            facebook_botkit.log('Got an unexpected message from Facebook: ', facebook_message);
                         }
                     }
                 }


libcthorne:
Same issue. These do indeed look like read receipt events:

```info: Got an unexpected message from Facebook:  { sender: { id: '**********************' },
  recipient: { id: '****************' },
  timestamp: 1465374667729,
  read: { watermark: 1465374667165, seq: 2896 } }```

anonrig:
Duplicate PR. Already, merged the other one.

Thanks for your contribution!

PR #266, opened 2016-06-08
Update Facebook.js

TimBroddin:
I saw some errors in my heroku logs. This should fix it.

Fix error "ReferenceError: botkit is not defined"


anonrig:
Thanks!

PR #278, opened 2016-06-15
Promisify the Slack Web API

mjesuele:
This pull request uses Bluebird to add promise support to the Slack Web API while retaining support for the original callback style.

If a callback is passed as the last argument to a Slack API call then it is executed as before. Otherwise, a promise is returned.

Feedback is very welcome. If there's interest, I can try my hand at adding promise support to the Facebook and/or Twilio interfaces as well.


mjesuele:
Anyone have any interest in merging this?

colestrode:
I'd like to hear some more discussion from the community. In other projects I've found maintaining both promises and callbacks gets pretty tricky (especially if there are multiple error states or early returns). Personally my favored approach is to support callbacks, since that's idiomatic for Node, and then users can promisify the API using bluebird or Q.

mjesuele:
@colestrode, bear with me if I'm missing something about your response, but I'm actually just using Bluebird's promisify method (https://github.com/howdyai/botkit/pull/278/commits/e8520ac2275a8bcd74d27db7fb61ccd84ef06e28#diff-e55a35b8c5887b6a2e0f96549d9a053fR42). If that's the approach you're suggesting the end user take (which I initially tried but couldn't figure out a reasonable way to do) then I'm not sure what the issue is with incorporating it directly into Botkit.

If you know of a way that I can promisify the API on my end, let me know, because then regardless of whether this PR gets merged, I can resume using the official Botkit package rather than my own fork as I'm doing now.

colestrode:
Hey @mjesuele :) Yeah, what I'm proposing is that the end user create their own promsified version of the lib. I did something similar with the storage module: https://github.com/colestrode/botkit-promise-storage (though I used Q instead of bluebird).

I have two concerns: the first is that, IMO, promisifying the slack api will create an inconsistent user experience since the rest of the botkit library only supports callbacks. I love promises, so if that's the way the community wants to go, I think we should go all in and support a promise-based interface. But having promises only in some places but not others leads to weird expectations and pitfalls for future users.

My second concern is around maintenance. In previous libraries I've contributed to we've attempted to support both promises and callbacks and it lead to a lot of complex code that wasn't easy to abstract, particularly around error handling or if there were multiple return points.

Again, I dig promises so I think this is really cool work and am really eager to hear what other people think. I just want to make sure we're thoughtful whatever we do :)

mjesuele:
@colestrode, I understand now about building a promisified version as a separate module, thanks for the link 😎

I see your point regarding the inconsistent user experience. Based on the number of thumbs-up this PR has gotten, I assume that there's a lot of interest in promisifying more of Botkit, and I'd be happy to contribute to that effort. I can see why we'd want to avoid introducing promise support in a piecemeal fashion -- at the same time, this is useful _now_.

It actually looks like the Twilio API is already promise-based, incidentally, and the Facebook adapter doesn't have any comparable API, so adding promise support to the Slack API might be a move towards greater internal coherence after all, even if the rest of Botkit remains callback-based.

I'm curious what the issue with maintenance is, at least for the simple case of the Slack API. My impression was that as long as the callbacks are indeed idiomatic Node, Bluebird's promisification should work pretty much automatically. With this last point, I might be asking for a little schooling, so I appreciate your time in responding.

Thanks!

colestrode:
Good point about the Twilio API already supporting promises! I didn't realize that, in that case I take back what I said about inconsistencies :) Promisfying the Slack API would bring more consistency.

As for maintenance, I like your approach of writing a "private" callback-style method and then conditionally promisfying that. That definitely makes the early exit and error cases easier to manage.

I think that would introduce some maintenance overhead if applied to the wider botkit framework, but honestly I'm not as concerned as I was :) So I'm for this if other people are :)

mjesuele:
@benbrown, is there anything addressable in the way of this being merged?

tlvenn:
Any update on this ?

ohbarye:
to maintainers
This PR seems not to break compatibility and to give a great feature which also people (including me) really want.
What's getting on the way of merge?

GauSim:
+1

tlvenn:
@benbrown any update beside the do not merge label plz ?

PR #298, opened 2016-06-24
Slack reconnect cleanup

jayjanssen:
Fixed some observed issues where reconnections would hit a ping/pong error right away.  This was due to the interval and the lastPong from the previous connection not being properly cleaned up.


benbrown:
thanks!!

PR #299, opened 2016-06-24
Show winston levels being set

jayjanssen:
I discovered a ton of botkit messages that weren't logging with the default levels in Winston, hence an update to the readme logger example.


benbrown:
thanks @jayjanssen

PR #302, opened 2016-06-25
reply to stickers with text on messenger

webtaculars:
Added a reply when a user sends stickers


luisfarzati:
This should be optional, as in a configuration param. I might want to get those emojis or whatever images.

anonrig:
Yes @luisfarzati you're right. Declining this PR because of this request.

PR #303, opened 2016-06-29
Basic Dockerfile to setup botkit

nguse:
A Dockerfile if you are interested in including it with this project.

This allows a user to easily setup and test/run this project using docker.

How a user could use this dockerfile (without docker hub):
```
git clone https://github.com/howdyai/botkit.git
docker build --tag botkit botkit/

Foreground:
docker run -it --rm --name botkit botkit sh -c 'token=<slack token> node slack_bot.js'

Background:
docker run -d --name botkit botkit sh -c 'token=<slack token> node slack_bot.js'
```

Users could easily specify their own js file from their host with something like:
```
docker run -d --volume=/home/foo/slack_bot.js:/app/slack_bot.js --name botkit botkit sh -c 'token=<slack token> node slack_bot.js'
```

My suggestion would be to setup an account on docker hub also so that you can leverage it to automatically build images for you and publish them for others to use.
https://hub.docker.com/

If this was setup on docker hub, users would be able to run this project with one command:
```
docker run -d --name botkit howdyai/botkit:latest sh -c 'token=<slack token> node slack_bot.js'
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,8 @@
+FROM library/node:slim
+
+COPY . /app
+
+RUN cd /app \
+  && npm install --production
+
+WORKDIR /app


anonrig:
Nice catch!

PR #100, opened 2016-01-24
Adds bot.replyIntent()

RafaelCosman:
Usage:

`bot.replyIntent(message, "greeting");`

Will hit `http://www.chatamize.com/api/v1/intent?api_key=[your api key]&intent=greeting` and will reply with the message returned. See `chatamize.com` to manage your messages.

As of `73374c0`, `intents-chatamize` passes all tests with `npm test`.
```
Test
    ✓ should have a token
    ✓ should have Botkit instance

  Botkit
info: ** No persistent storage method specified! Data may be lost when process shuts down.
info: ** Setting up custom handlers for processing Slack messages
info: ** API CALL: https://slack.com/api/rtm.start
notice: ** BOT ID:  intent-wit-bot  ...attempting to connect to RTM!
    ✓ should start and then stop (1782ms)
info: ** No persistent storage method specified! Data may be lost when process shuts down.
info: ** Setting up custom handlers for processing Slack messages
info: ** API CALL: https://slack.com/api/rtm.start
    ✓ should have fail with false token (171ms)

  Log
    ✓ should use an external logging provider (177ms)


  5 passing (2s)
```

PR #306, opened 2016-07-01
Add support for Quick Reply as per Messenger v1.1 API

MironV:
This adds the ability both send and receive quick_reply payloads.


anonrig:
Seems good!

mvaragnat:
thank you !

asveloper:
@MironV i am new to botkit can you please provide the format of the object in which data should be sent for quick reply. I am getting this error 

API ERROR { message: '(#100) Must send either message or state'

PR #317, opened 2016-07-10
Facebook POST message using JSON and newer endpoint to support Emojis.

tkornblit:
The older Facebook endpoint doesn't support (All) Emojis. Using JSON and the new endpoint theres now support :).

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -823,7 +823,7 @@ function Botkit(configuration) {
             // set up a once a second tick to process messages
             botkit.tickInterval = setInterval(function() {
                 botkit.tick();
-            }, 1000);
+            }, 1500);
         }
     };
 
--- a/lib/Facebook.js
+++ b/lib/Facebook.js
@@ -47,32 +47,35 @@ function Facebookbot(configuration) {
                 facebook_message.message.quick_replies = message.quick_replies;
             }
 
-            request.post('https://graph.facebook.com/me/messages?access_token=' + configuration.access_token,
+            //Add Access Token to outgoing request
+
+            facebook_message.access_token = configuration.access_token;
+
+            request({
+                method: "POST",
+                json: true,
+                headers: {
+                    "content-type": "application/json",
+                },
+                body: facebook_message,
+                uri: 'https://graph.facebook.com/v2.6/me/messages'
+            },
                 function(err, res, body) {
+
+
                     if (err) {
                         botkit.debug('WEBHOOK ERROR', err);
                         return cb && cb(err);
                     }
 
-                    try {
-
-                        var json = JSON.parse(body);
-
-                    } catch (err) {
-
-                        botkit.debug('JSON Parse error: ', err);
-                        return cb && cb(err);
-
-                    }
-
-                    if (json.error) {
+                    if (body.error) {
                         botkit.debug('API ERROR', json.error);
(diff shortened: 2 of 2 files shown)


krismuniz:
So, now you can specify the `access_token` in the body of the request instead of a query string in the url?

tkornblit:
I think the problem was more in the Form be JSON way of posting to Facebook, and potentially with the endpoint version itself.

The access token can be in the URL or body. Doesn't matter.

dcsan:
hey, even with the old version (current botkit) how are you sending emoji to facebook?
are you using some type of pseudo codes? or just the basic emoji FB will convert ( 3> etc)

tkornblit:
Using the new update you can send them with plaintext. No pseudo code required.

dcsan:
right, but waiting for that to get merged...​ meantime wanted to know what
pseudo code methods there are.

tkornblit:
Not sure I understand the question

dcsan:
until your PR is merged - how do you send emoji to FBM?

tkornblit:
Lol, you can't.

lib/CoreBot.js
```diff
@@ -823,7 +823,7 @@ function Botkit(configuration) {
             // set up a once a second tick to process messages
             botkit.tickInterval = setInterval(function() {
                 botkit.tick();
-            }, 1000);
+            }, 1500);
```
anonrig: Is there a particular reason to change the setInterval method to 1500 ms?

lib/Facebook.js
```diff
@@ -47,32 +47,35 @@ function Facebookbot(configuration) {
                 facebook_message.message.quick_replies = message.quick_replies;
             }
 
-            request.post('https://graph.facebook.com/me/messages?access_token=' + configuration.access_token,
+            //Add Access Token to outgoing request
```
anonrig: Please use JSDoc for documentation purposes.

lib/CoreBot.js
tkornblit: Turns out you can revert this as I don’t think it makes a difference. Basically the issue with the framework with Facebook is that if you send 2 or 3 messages, and one of them is an image for some reason it will come out of order. It’s something with facebooks queue on the backend. So i think what needs to happen is to wait in between messages before sending
--
Tom Kornblit

dcsan:
this is a month old, would be nice to get merged in, a lot of chatter in #fb-messenger about emoji.

tkornblit:
What do you need from me to get this merged?

benbrown:
So sorry for the delay! Merging. Thank you for this fix.

tkornblit:
Awesome. This actually comes at an interesting time. I wanted to run something by you and see what you thought.

The past 4 months I’ve been working on building my own framework that wraps botkit. It essentially exposes a Data Description JSON Language that enables you to write bots using just data.

I think this will be really helpful for other developers, so I’m open-sourcing it, in hopes that it helps developers build more rapidly and cross platform.

I call the project Mermaid <https://github.com/sage-bots/mermaid>, and would love to hear your thoughts.

-Tom
--
Tom Kornblit
Founder of Sage <http://www.sagebots.com/>
Personal Blog <http://blog.tomkornblit.com/> | LinkedIn <https://www.linkedin.com/in/tomkornblit> | Twitter <https://twitter.com/tkornblit>

PR #319, opened 2016-07-11
simplifies slack web api instantiation

colestrode:
This change simplifies the creation of the Slack web api interface and reduces the amount of copy paste code.

I also simplified the callbacks in the callAPI and callAPIWithoutToken methods.

### DIFF
--- a/lib/Slack_web_api.js
+++ b/lib/Slack_web_api.js
@@ -95,7 +95,7 @@ module.exports = function(bot, config) {
                 try {
                     json = JSON.parse(body);
                 } catch (err) {
-                    return cb(err || 'Invalid JSON');
+                    return cb(err);
                 }
 
                 if (json.ok) {
--- a/lib/Slack_web_api.js
+++ b/lib/Slack_web_api.js
@@ -152,7 +152,12 @@ module.exports = function(bot, config) {
     // overwrite default behavior
     slack_api.chat.postMessage = function(options, cb) {
         if (options.attachments && typeof(options.attachments) != 'string') {
-            options.attachments = JSON.stringify(options.attachments);
+            try {
+                options.attachments = JSON.stringify(options.attachments);
+            } catch(err) {
+                delete options.attachments;
+                bot.log.error('Could not parse attachments', err);
+            }
         }
         slack_api.callAPI('chat.postMessage', options, cb);
     };
--- a/lib/Slack_web_api.js
+++ b/lib/Slack_web_api.js
@@ -114,15 +114,13 @@ module.exports = function(bot, config) {
         var groups = slackMethod.split('.');
         var method = groups.pop();
         var currentGroup = slack_api;
-        var nextGroupName;
 
-        for (var i = 0; i < groups.length; i++) {
-            nextGroupName = groups[i];
+        groups.forEach(function(nextGroupName) {
             if (!currentGroup[nextGroupName]) {
                 currentGroup[nextGroupName] = {};
             }
             currentGroup = currentGroup[nextGroupName];
-        }
+        });
 
         currentGroup[method] = function(options, cb) {
             slack_api.callAPI(slackMethod, options, cb);
--- a/lib/Slack_web_api.js
+++ b/lib/Slack_web_api.js
@@ -1,314 +1,180 @@
 var request = require('request');
 
+function noop() {}
+
 module.exports = function(bot, config) {
-
-    // create a nice wrapper for the Slack API
     var slack_api = {
-        api_url: 'https://slack.com/api/',
-        // this is a simple function used to call the slack web API
-        callAPI: function(command, options, cb) {
-            bot.log('** API CALL: ' + slack_api.api_url + command);
-            if (!options.token) {
-                options.token = config.token;
-            }
-            bot.debug(command, options);
-            request.post(this.api_url + command, function(error, response, body) {
-                bot.debug('Got response', error, body);
-                if (!error && response.statusCode == 200) {
-                    var json;
-                    try {
-                        json = JSON.parse(body);
-                    } catch (err) {
-                        if (cb) return cb(err || 'Invalid JSON');
-                        return;
-                    }
+        api_url: 'https://slack.com/api/'
+    };
 
-                    if (json.ok) {
-                        if (cb) cb(null, json);
-                    } else {
-                        if (cb) cb(json.error, json);
-                    }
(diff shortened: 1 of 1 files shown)


lib/Slack_web_api.js
```diff
@@ -1,314 +1,180 @@
...
-            remove: function(options, cb) {
-                slack_api.callAPI('pins.remove', options, cb);
+            return cb(error || 'Invalid response');
+        }).form(options);
+    };
+
+
+    // generate all API methods
+    slackApiMethods.forEach(function(slackMethod) {
+        // most slack api methods are only two parts, of the form group.method, e.g. auth.test
```
anonrig: All comments should be written using JSDoc. Can you update all of these comments, please?

lib/Slack_web_api.js
```diff
@@ -1,314 +1,180 @@
...
+            options.redirect_uri = bot.config.redirectUri;
+        }
+        request.post(slack_api.api_url + command, function(error, response, body) {
+            bot.debug('Got response', error, body);
+            if (!error && response.statusCode == 200) {
+                var json;
+                try {
+                    json = JSON.parse(body);
+                } catch (err) {
+                    return cb(err || 'Invalid JSON');
```
anonrig: `err` is an instance of `Error` class but if `err` doesn't exist, and the body is an invalid JSON you're sending `String`, therefore it makes the compatibility much harder to comprehend.

lib/Slack_web_api.js
```diff
@@ -1,314 +1,180 @@
...
+        // some have three parts: group.subgroup.method, e.g, users.profile.get
+        // this method loops through all groups in a method, ensures they exist,
+        // then adds the method to the terminal group
+
+        var groups = slackMethod.split('.');
+        var method = groups.pop();
+        var currentGroup = slack_api;
+        var nextGroupName;
+
+        for (var i = 0; i < groups.length; i++) {
```
anonrig: Using groups.length in a for loop will calculate the length of `groups` array in every iteration. (Because JavaScript sucks.) Please use lodash, `Array.forEach` or calculate the length of groups array before the loop.

lib/Slack_web_api.js
```diff
@@ -1,314 +1,180 @@
...
+        currentGroup[method] = function(options, cb) {
+            slack_api.callAPI(slackMethod, options, cb);
+        };
+
+    });
+
+    // overwrite default behavior
+    slack_api.chat.postMessage = function(options, cb) {
+        if (options.attachments && typeof(options.attachments) != 'string') {
+            options.attachments = JSON.stringify(options.attachments);
```
anonrig: `JSON.stringify` may throw an exception called `TypeError: Converting circular structure to JSON`, therefore, can you update this part of the code to be more defensive?

lib/Slack_web_api.js
```diff
@@ -1,314 +1,180 @@
 var request = require('request');
 
-module.exports = function(bot, config) {
+function noop() {}
```
anonrig: Do you have a particular reason to call this function `noop`?

lib/Slack_web_api.js
colestrode: Sure thing, is there a jsdoc convention for inline docs like this? Most of what I know is for method signatures or class definitions.

lib/Slack_web_api.js
colestrode: It looks like array.forEach is not supported before Node 4.0.0 (http://node.green/) Is it ok to add lodash as a dependency or should just cache the length of the array? Fwiw, lodash will probably come in handy in other places.

lib/Slack_web_api.js
colestrode: Alrighty; what should be the behavior in case JSON parsing fails? Seems that the method should fail, maybe I can throw a more friendly error message? On the other hand, it does seems more informative to let the raw error bubble out, since that will be actionable by the developer.

lib/Slack_web_api.js
colestrode: Good catch, that `||` isn't really doing anything, I'll just take it out.

lib/Slack_web_api.js
colestrode: Yeah, `noop` is short for "no operation", and has become a convention for methods that don't do anything (https://en.m.wikipedia.org/wiki/NOP)

lib/Slack_web_api.js
```diff
@@ -1,314 +1,172 @@
...
     return slack_api;
 
+
+    /**
+     * Makes a POST request as a form to the given url with the options as data
+     * @param url
+     * @param options
+     * @param cb
+     */
+    function postForm(url, options, cb) {
```
colestrode: I also pulled out this request.post and any other duplicate code, this reduces callAPI and callAPIWithoutToken to constructing options objects, all other request logic is handled here

lib/Slack_web_api.js
anonrig: The possibility of having a JSON.stringify exception is really low to implement a retry method, therefore, I believe silently (with giving an error message to the console with the reason) would be a more appropriate solution.

lib/Slack_web_api.js
anonrig: Having a function and removing the code duplicate parts is a good choice!

lib/Slack_web_api.js
colestrode: I think I got all the JSDoc ironed out 😄

lib/Slack_web_api.js
```diff
@@ -1,314 +1,196 @@
...
+
+        bot.log('** API CALL: ' + url);
+        request.post(url, function(error, response, body) {
+            bot.debug('Got response', error, body);
+            if (!error && response.statusCode == 200) {
+                var json;
+                try {
+                    json = JSON.parse(body);
+                } catch (parseError) {
+                    return cb(parseError);
```
anonrig: You've indicated that callback is optional. But you didn't check if cb is defined or not. Is it truely optional?

colestrode: The first line checks for a callback and if it's not defined sets it to the noop function. That simplifies the checks later since cb will always be defined.

benbrown:
@colestrode sorry it took so long to merge this!
thanks for cleaning everything up so nicely.

colestrode:
No problem @benbrown glad to help 😀

PR #320, opened 2016-07-12
Added FB sticker detection

MironV:
FB stickers can be distinguished from regular images sent by the user by looking at the message data for a sticker_id field. Right now, botkit doesn't support plumbing that through so it's impossible to distinguish a Like from someone sending an image. This fixes that issue.

### DIFF
--- a/lib/Facebook.js
+++ b/lib/Facebook.js
@@ -140,7 +140,7 @@ function Facebookbot(configuration) {
                                 timestamp: facebook_message.timestamp,
                                 seq: facebook_message.message.seq,
                                 mid: facebook_message.message.mid,
-								sticker_id: facebook_message.message.sticker_id,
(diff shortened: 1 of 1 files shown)


lib/Facebook.js
```diff
@@ -136,6 +140,7 @@ function Facebookbot(configuration) {
                                 timestamp: facebook_message.timestamp,
                                 seq: facebook_message.message.seq,
                                 mid: facebook_message.message.mid,
+								sticker_id: facebook_message.message.sticker_id,
```
anonrig: There is an indentation difference in here. Can you solve it, please? Thanks.

MironV: That should fix it!

benbrown:
Cool!

Should this maybe also throw a special event, like sticker_received?

PR #323, opened 2016-07-13
Fix #159 - Add support for ip on webserver

### LINKED ISSUE #159: Setup Webserver on RedHat OpenShift

LoneRifle:
This PR is based on - and supersedes - #249, and addresses an issue with that PR, namely, the quoting of the localhost IP as a string. It adds a new supported field - ip - to the config consumed by the Slack variant of Botkit so that it can be used out-of-the-box on hosting providers (like RedHat Openshift) that require processes to listen on specific IP addresses.

This issue fixes #159.

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -207,9 +207,13 @@ function Slackbot(configuration) {
 
         if (slack_botkit.config && slack_botkit.config.webserver && slack_botkit.config.webserver.static_dir)
             static_dir = slack_botkit.config.webserver.static_dir;
-
-        slack_botkit.config.port = port;
-        slack_botkit.config.ip = slack_botkit.config.ip || '127.0.0.1';
+        
+        if (slack_botkit.config) {
+            slack_botkit.config.port = port;
+            slack_botkit.config.ip = slack_botkit.config.ip || '127.0.0.1';
+        }
+        
+        var ip = slack_botkit.config? slack_botkit.config.ip : '127.0.0.1';
 
         slack_botkit.webserver = express();
         slack_botkit.webserver.use(bodyParser.json());
@@ -217,12 +221,10 @@ function Slackbot(configuration) {
         slack_botkit.webserver.use(express.static(static_dir));
 
         var server = slack_botkit.webserver.listen(
-            slack_botkit.config.port,
-            slack_botkit.config.ip,
+            port,
+            ip,
             function() {
-                slack_botkit.log('** Starting webserver on ' +
-                    ((slack_botkit.config.ip)?slack_botkit.config.ip+':':'port ') +
-                    slack_botkit.config.port);
+                slack_botkit.log('** Starting webserver on ' + ip + ':' + port);
                 if (cb) { cb(null, slack_botkit.webserver); }
             });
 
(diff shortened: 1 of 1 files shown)


lib/SlackBot.js
```diff
@@ -217,8 +218,10 @@ function Slackbot(configuration) {
 
         var server = slack_botkit.webserver.listen(
             slack_botkit.config.port,
+            slack_botkit.config.ip,
             function() {
-                slack_botkit.log('** Starting webserver on port ' +
+                slack_botkit.log('** Starting webserver on ' +
+                    ((slack_botkit.config.ip)?slack_botkit.config.ip+':':'port ') +
```
anonrig: Can you make this line more readable please?

lib/SlackBot.js
```diff
@@ -209,6 +209,7 @@ function Slackbot(configuration) {
             static_dir = slack_botkit.config.webserver.static_dir;
 
         slack_botkit.config.port = port;
+        slack_botkit.config.ip = slack_botkit.config.ip || '127.0.0.1';
```
anonrig: Can you double check if `slack_botkit.config` exists. If it's not, it will crash eventually.

LoneRifle:
Changes applied in response to comments, please review.

lib/SlackBot.js
```diff
@@ -207,19 +207,24 @@ function Slackbot(configuration) {
...
+        var ip = slack_botkit.config? slack_botkit.config.ip : '127.0.0.1';
 
         slack_botkit.webserver = express();
         slack_botkit.webserver.use(bodyParser.json());
         slack_botkit.webserver.use(bodyParser.urlencoded({ extended: true }));
         slack_botkit.webserver.use(express.static(static_dir));
 
         var server = slack_botkit.webserver.listen(
-            slack_botkit.config.port,
+            port,
```
anonrig: Why did you change this line to port?

If config exists port is not changed. But, slack_botkit.config.port changes to port. Therefore, there exist a logical error.

LoneRifle: Can you clarify or rephrase your sentence please? You imply in an earlier diff comment that we cannot assume that slack_botkit.config is defined. Given this, and the fact that we enforce that port is defined earlier in lines 199-201 and will overwrite anything in slack_botkit.config.port, the use of port should not result in any change in behaviour.

LoneRifle:
Change of port argument done as requested, please review.

LoneRifle:
Could I have an update on this PR please?

Issue #307, opened 2016-07-05
slack_botkit not defined in middleware


gDelgado14:
Using Botkit 0.2.2 on Mac OSX el Capitan

My server throws a ReferenceError when trying to do custom slash commands. 

The error points to `slack_authentication.js` where there is no reference made to slack_botkit

```javascript
slack_botkit.log(
            '** Requiring token authentication for webhook endpoints for Slash commands ' +
            'and outgoing webhooks; configured ' + tokens.length + ' tokens'
        );
```

frmendes:
+1.

```js
var controller = Botkit.slackbot(config);
var bot = controller.spawn({
  incoming_webhook: {
    url: process.env.INCOMING_WEBHOOK_URL
  }
});

controller.setupWebserver(process.env.PORT, function(err, express_webserver) {
  controller.createWebhookEndpoints(express_webserver,
      [process.env.OUTGOING_WEBHOOK_TOKEN]);
});
```

At this point, it gives the same `ReferenceError`:

```
ReferenceError: slack_botkit is not defined
    at authenticate (/path/to/project/node_modules/botkit/lib/middleware/slack_authentication.js:39:9)
    at Layer.handle [as handle_request] (/path/to/project/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (/path/to/project/node_modules/express/lib/router/index.js:312:13)
    at /path/to/project/node_modules/express/lib/router/index.js:280:7
    at Function.process_params (/path/to/project/node_modules/express/lib/router/index.js:330:12)
    at next (/path/to/project/node_modules/express/lib/router/index.js:271:10)
    at serveStatic (/path/to/project/node_modules/serve-static/index.js:75:16)
    at Layer.handle [as handle_request] (/path/to/project/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (/path/to/project/node_modules/express/lib/router/index.js:312:13)
    at /path/to/project/node_modules/express/lib/router/index.js:280:7
```

Downgrading to `0.2.1` seems to fix the issue.

sundeepgupta:
It appears safe to comment out that logging code, I've notified Howdy folks about it.

sundeepgupta:
BTW downgrading to 0.2.1 will break the token check when the slash command POST comes in to your server.

peterswimm:
PR related to this issue has been released.

PR #328, opened 2016-07-17
Issue 307: Fix reference error in slash command processing.

sundeepgupta:
https://github.com/howdyai/botkit/issues/307

There is logging happening here which introduces a reference error on `slack_botkit`. Removed this logging code because it already happens in SlackBot.js.


anonrig:
Thanks!

Issue #329, opened 2016-07-17
Pass Slack OAuth `redirect_uri` query params when present


sundeepgupta:
The use case is a user issues a slash command, but we can't process their slash command because we don't have their access token yet. So we save the command in our server and respond with an auth link. The user goes through the auth flow and now that we have a token, we wan't to process their original command. 

Before Botkit, I was doing this be adding a `command=99` query param onto the `redirect_uri` and I'd check for it when Slack POSTed back with the user's token. It worked nicely.

With Botkit, It looks like I'd have to manually process the POST request. Perhaps this could be passed into the `create_user` event?

sundeepgupta:
@peterswimm any thoughts on this? Our app needs a solution to handle oauth flows in different ways, right now I'll have to use a hard fork. Wondering if you see the value in this. I'd imagine all Slack apps would want to behave this way, but maybe it's only me?

PR #343, opened 2016-07-25
Add flexibility to Slack auth flow via custom redirect params

sundeepgupta:
https://github.com/howdyai/botkit/issues/329

This will enable handling auth flows for various use cases by way of `redirect_uri` query params. 

If we stay with the existing design, I see two approaches:
a) Pass params around and only use them if the Slack App controller is configured with the `redirectUri` value.
b) Pass the entire redirect URI around. This options gives more flexibility in case apps want to have multiple redirect URI endpoints. This might be useful if they want greater control over their auth flows than what Botkit provides.

For the app we're building, we don't have a use case for b), so I thought to try out a) and get some feedback. 

Question (if the approach is ok):
Should the `redirect_params` be passed into the `create/update_team` and `create_bot` events also?

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -2,6 +2,7 @@ var Botkit = require(__dirname + '/CoreBot.js');
 var request = require('request');
 var express = require('express');
 var bodyParser = require('body-parser');
+var querystring = require('querystring');
 
 function Slackbot(configuration) {
 
@@ -303,7 +304,7 @@ function Slackbot(configuration) {
     };
 
     // get a team url to redirect the user through oauth process
-    slack_botkit.getAuthorizeURL = function(team_id) {
+    slack_botkit.getAuthorizeURL = function(team_id, redirect_params) {
 
         var scopes = slack_botkit.config.scopes;
         var url = 'https://slack.com/oauth/authorize' + '?client_id=' +
@@ -312,8 +313,14 @@ function Slackbot(configuration) {
         if (team_id)
             url += '&team=' + team_id;
 
-        if (slack_botkit.config.redirectUri)
-            url += '&redirect_uri=' + slack_botkit.config.redirectUri;
+        if (slack_botkit.config.redirectUri) {
+            var redirect_query = '';
+            if (redirect_params)
+                redirect_query += encodeURIComponent(querystring.stringify(redirect_params));
+
+            var redirect_uri = slack_botkit.config.redirectUri + '?' + redirect_query;
+            url += '&redirect_uri=' + redirect_uri;
+        }
 
         return url;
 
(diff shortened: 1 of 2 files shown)


sundeepgupta:
bump, I welcome any comments/questions. Or better yet, a merge if this is good as is :)

sundeepgupta:
Hey Botkit folks, I welcome any comments/feedback on this. We're using this (via our fork) quite successfully so far.

lib/SlackBot.js
```diff
@@ -382,7 +389,17 @@ function Slackbot(configuration) {
                 code: code
             };
 
-            if (slack_botkit.config.redirectUri) opts.redirect_uri = slack_botkit.config.redirectUri;
+            var redirect_params = {};
+            if (slack_botkit.config.redirectUri) {
+              Object.assign(redirect_params, req.query);
+              delete redirect_params.code
```
anonrig: missing `;`

lib/SlackBot.js
```diff
@@ -382,7 +389,17 @@ function Slackbot(configuration) {
...
+              delete redirect_params.state
```
anonrig: missing `;`

lib/SlackBot.js
```diff
@@ -382,7 +389,17 @@ function Slackbot(configuration) {
                 code: code
             };
 
-            if (slack_botkit.config.redirectUri) opts.redirect_uri = slack_botkit.config.redirectUri;
+            var redirect_params = {};
+            if (slack_botkit.config.redirectUri) {
```
anonrig: Can you add redirectUri to DOCUMENTATION?

lib/SlackBot.js
```diff
@@ -2,6 +2,7 @@ var Botkit = require(__dirname + '/CoreBot.js');
 var request = require('request');
 var express = require('express');
 var bodyParser = require('body-parser');
+var querystring = require('querystring');
```
anonrig: you should add querystring to package.json

sundeepgupta: Will do. As a new Node.js developer, can you help me understand why this needs to be in the package.json?

lib/SlackBot.js
sundeepgupta: Will do.

lib/SlackBot.js
anonrig: if you're using https://www.npmjs.com/package/querystring querystring npm module, you should install the package with `npm install querystring --save` to save the node module name to package.json. Additionally, if you don't do it, and next time a developer installs the app using `npm install` the querystring node module will not be installed and therefore result in a crash.

sundeepgupta:
@anonrig Great thanks for explaining. I've updated the branch to address the feedback.

lib/SlackBot.js
anonrig: @sundeepgupta I believe you were using the built in querystring instead of NPM module. You don't need the package.json modification for that.

sundeepgupta: @anonrig ok great, pushed up new version

sundeepgupta:
@anonrig quick bump to see if anything else was needed. I think I've addressed everything, but otherwise, please let me know... thanks.

peterswimm:
@sundeepgupta sorry for the delay, we'll get on evaluating this and other pending PRs very shortly.

sundeepgupta:
Awesome thanks! 

Cheers,
Sundeep

sundeepgupta:
Thought I'd give this another bump...

sundeepgupta:
:nudge:

josephfinlayson:
Please merge!

benbrown:
AT LONG LAST, merged!

Thanks for your patience, @sundeepgupta

sundeepgupta:
yay!

Issue #156, opened 2016-03-19
No way to get message meta from RTM send?


dclowd9901:
I want to be able to update a message after sending it using the api, but the api requires a precision time stamp and channel to update the message. As far as I can tell, the only instance in which we get usable metadata from botkit is when slackbot "hears" an incoming message. So if slackbot posts a message, that metadata is not retrieved, and that message cannot be updated. Am I missing something?

dclowd9901:
Even when digging deep into the bowels of Slackbot_worker.js to line 254, and console logging out the arguments, the arguments object appears to be completely empty if the request is good. Is this a shortcoming of websockets/rtm?

peterswimm:
Fixed in https://github.com/howdyai/botkit/pull/344

PR #344, opened 2016-07-27
Add replyAndUpdate method for updating Slack messages

declandewet:
This pull request adds a new method to `Slackbot_worker.js`: `replyAndUpdate` which allows quick editing of a previously sent reply. 

Which addresses this issue: #156 

(I also snuck in a commit that fixes some linter issues outside of this implementation)

Here is a GIF demonstrating it in action:

botkit-update

I figured that there was so much boilerplate code necessary when updating previously sent messages that it warranted a helper function, and this is what I've come up with.

It takes a `source` message, a `response` message (string or object) (the "_updateable_" - that is, a message which may or may not be updated) and an asynchronous "_updater_" callback which determines how and when to update the "_updateable_" `response` message. Internally, `replyAndUpdate` ensures that a response message with a valid `ts` timestamp is returned to the callback by fooling `bot.reply` into using the Slack Web API instead of RTM messaging - it does this by ensuring that there is always an `attachments` field on the `response` message.

The "_updater_" callback receives two arguments - an `error` (passed from `bot.reply` when sending the initial response message) and an `updateResponse` function. The `updateResponse` function accepts an "_update_" message (string or object) (which internally will receive the `ts` and `channel` properties of the `response` message) and an optional callback which receives an error from `bot.api.chat.update`.

Here are two examples of how it could be used (though contrived - you would never typically do the following in a real chatbot app - they serve as the simplest examples to get the point across without too much cognitive overhead):

```js
// fixing a typo
controller.hears('hello', ['ambient'], (bot, msg) => {
  // send a message back: "hellp"
  return bot.replyAndUpdate(msg, 'hellp', (error, updateResponse) => {
    if (error) return console.error(error);
    // oh no, "hellp" is a typo - let's update the message to "hello"
    return updateResponse('hello', (error) => console.error(error));
  });
});

// updating on result of third-party APIs
// (though, you should probably use the `typing` event for this instead of "Ok, let me think...")
controller.hears('tell me a joke', ['ambient'], (bot, msg) => {
  return bot.replyAndUpdate(msg, 'Ok, let me think...', (error, updateResponse) => {
    if (error) return console.error(error);
    return http.get('https://joke.api/random', (error, response) => {
      if (error) return updateResponse('Sorry, but I cannot think of a joke right now.', console.error.bind(console));
      let joke;
      try {
        joke = JSON.parse(response);
      } catch (error) {
        return updateResponse('Sorry, but I cannot think of a joke right now.', console.error.bind(console));
      }
      return updateResponse(joke.text, console.error.bind(console));
    });
  });
});
```

### DIFF
--- a/lib/Slackbot_worker.js
+++ b/lib/Slackbot_worker.js
@@ -570,6 +570,45 @@ module.exports = function(botkit, config) {
     };
 
     /**
+     * replies with message, performs arbitrary task, then updates reply message
+     * note: don't use this as a replacement for the `typing` event
+     *
+     * @param {Object} src - message source
+     * @param {(string|Object)} resp - response string or object
+     * @param {function} [cb] - updater callback
+     */
+    bot.replyAndUpdate = function(src, resp, cb) {
+        try {
+            resp = typeof resp === 'string' ? { text: resp } : resp;
+            // trick bot.reply into using web API instead of RTM
+            resp.attachments = resp.attachments || [];
+        } catch (err) {
+            return cb && cb(err);
+        }
+        // send the "updatable" message
+        return bot.reply(src, resp, function(err, src) {
+            if (err) return cb && cb(err);
+
+            // if provided, call the updater callback - it controls how and when to update the "updatable" message
+            return cb && cb(null, function(resp, cb) {
+                try {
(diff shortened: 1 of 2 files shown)


anonrig:
Perfect!

PR #353, opened 2016-08-02
don't require callback for synchronous identify functions

noam3127:
Perhaps this is just a personal preference, but I believe that callbacks should only be required for functions that may perform asynchronous operations. For all synchronous functions, requiring a callback to be passed in may mislead devs into thinking that it's asynchronous and requires unnecessary consideration for control flow. Therefore, I made the callbacks optional for `bot.identifyBot()` and `bot.identifyTeam()` because there's absolutely nothing async taking place in those.

### DIFF
--- a/lib/Slackbot_worker.js
+++ b/lib/Slackbot_worker.js
@@ -227,34 +227,39 @@ module.exports = function(botkit, config) {
     };
 
     bot.identifyBot = function(cb) {
+        var data;
         if (bot.identity) {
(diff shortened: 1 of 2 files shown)


anonrig:
It's a good approach when using `yield` on ES6, but if you're not using it, it gives a asynchronous feeling to a asynchronous function.

benbrown:
I will allow it! Thanks.

PR #358, opened 2016-08-08
Going through Slack OAuth flow should update the user's token and sco…

sundeepgupta:
…pes.

This is necessary because the OAuth access token could become invalid or revoked. Also the application's required scopes could change. If these things happen and we need to have the user auth again, the new tokens/scopes should be updated on that user.


Intellogo:
hello. when is this scheduled for release?

PR #360, opened 2016-08-09
fix single-character utterances

noam3127:
This simply will consider "n" as a "no" utterance and "y" as a "yes" utterance _only_ if there are no more characters in the string. This protects against conflicts between capturing-options, for instance, when one option matches a word like `'^next`, and another matches "no"utterances so when the user responds with "next", we don't want it to be considered a "no" utterance erroneously.

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -24,8 +24,8 @@ function Botkit(configuration) {
     };
 
     botkit.utterances = {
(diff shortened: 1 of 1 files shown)


YemSalat:
Sorry, I dont have full context, but do you think the "yeah" and "nope" checks are redundant, since they are 'covered' by "yea" and "no" respectively? I mean if it matches "yea" - it will also match "yea*h*" - wont it?

Also, perhaps should add tests for these.

anonrig:
Sounds good to me.

PR #361, opened 2016-08-10
Update command-line-args to new API version in facebook_bot.js

houjunchen:
According to the [release note](https://github.com/75lb/command-line-args/releases) of command-line-args, there is a breaking change between v2.1.6 and v3.0.0 which will cause runtime error.


benbrown:
Thank you.

PR #363, opened 2016-08-12
Fixed typo on logs, 'interactive message' instead of 'slash command'

deubaka:
Just something minor but bothered me during debugging/exploring.

Cheers,
Deu


benbrown:
Thank you!

PR #367, opened 2016-08-13
Add user_info to the bot object

rmd6502:
Adds a dictionary from the users key of the auth response, indexed by id, and containing the information about users who are members of the chat room.

See line 75 of [insultbot.js](https://github.com/rmd6502/insultbot/blob/master/insultbot.js) for an example.

screen shot 2016-08-13 at 2 02 30 pm

### DIFF
--- a/lib/Slackbot_worker.js
+++ b/lib/Slackbot_worker.js
@@ -131,10 +131,14 @@ module.exports = function(botkit, config) {
 
             bot.identity = res.self;
             bot.team_info = res.team;
+            bot.user_info = {};
+            res.users.forEach(userRecord => {
+                bot.user_info[userRecord.id] = userRecord;
+            });
 
             /**
              * Also available:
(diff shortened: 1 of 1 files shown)


lib/Slackbot_worker.js
```diff
@@ -131,10 +131,14 @@ module.exports = function(botkit, config) {
 
             bot.identity = res.self;
             bot.team_info = res.team;
+            bot.user_info = {};
+            res.users.forEach(userRecord => {
```
anonrig: What if `res.users` is undefined or not an array? Can you write this code defensively?

benbrown:
This is generally a good idea: the RTM payload contains tons of useful information that, if cached, would remove the need to call the API to look up the info.

HOWEVER, it opens a small can of worms because the information becomes out of date as users change their info, or are added or removed from the team.

It would be great to see this PR include handlers for that type of change event so that this cache could be kept up to date.

PR #368, opened 2016-08-13
Add getUserlist function

alexdorn87:
Allow to get user list from slack api.

### DIFF
--- a/lib/Slackbot_worker.js
+++ b/lib/Slackbot_worker.js
@@ -114,7 +114,11 @@ module.exports = function(botkit, config) {
         bot.closeRTM();
         botkit.shutdown();
     };
-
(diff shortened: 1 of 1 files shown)


lib/Slackbot_worker.js
```diff
@@ -114,7 +114,11 @@ module.exports = function(botkit, config) {
         bot.closeRTM();
         botkit.shutdown();
     };
-
+    
+    bot.getUserList = function (token, cb) {
```
anonrig: Can you add a JsDoc for this function?

benbrown:
The best way to do this is to directly call bot.api.users.list()

You should definitely _not_ pass in a different token here.

PR #370, opened 2016-08-14
Maintain title length to 20 character in quick reply

webtaculars:
If the title length of a quick reply element in facebook messenger is more than 20 character then it shows a error. So I have added a function to keep the title in limit.

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -228,8 +228,31 @@ function Botkit(configuration) {
 
 
         this.ask = function(message, cb, capture_options) {
+          if (message.quick_replies !== undefined) {
+            var newMessage = {
+              text: message.text,
+              quick_replies: message.quick_replies.map(function(item) {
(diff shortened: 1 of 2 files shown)


lib/CoreBot.js
```diff
@@ -228,8 +228,31 @@ function Botkit(configuration) {
 
 
         this.ask = function(message, cb, capture_options) {
+          if (message.quick_replies !== undefined) {
```
anonrig: undefined should be string like 'undefined'

webtaculars: No. It should be undefined. When a message has quick replies, message.quick_replies will not be undefined. When a message does not have quick replies, message.quick_replies will undefined. It won't be a string

benbrown:
I implemented this change in a slightly different way place -- inside the Facebook send command.

Thanks for contributing!

PR #386, opened 2016-08-25
Fix authentication for Slack Interactive Messages.

sundeepgupta:
The incoming request for an Interactive Message has a different structure than Slash Commands. The authentication middleware should be able to handle this different format. Specifically, the request body has a `payload` JSON string, which contains the token.


anonrig:
Thank you for your contribution @sundeepgupta

jfairley:
I need this for a slash_command in my bot. Any chance of getting a new push to NPM soon?

Issue #29, opened 2015-12-20
Fix file uploading


amacgregor:
I can't get the file upload functionality working this is my latest attempt:

```
            var filepath = path.join(__dirname, '..', '/screenshots/'+ name + '.png');
            fs.readFile(filepath, function (err, data) {
                if (err) throw err;
                
                console.log(message.channel);
                
                bot.api.files.upload({
                    content: data,
                    channel: message.channel
                },function(err,res) {
                    if (err) {
                        bot.botkit.log("Failed to add screenshot :(",err);
                        bot.botkit.log(data);
                    }
                });
```

But I don't get anything back, the data is there I also tried replacing content: with file and I get an error from the api saying not file data. Any help is highly appreciated.

ZuzooVn:
@amacgregor : Did you find the solution?

demac-rd:
I could never make the botkit method work so I ended up using a curl wrapper for the fileuploads that worked like a charm:

```
            var filepath = path.join(__dirname, '..', '/screenshots/'+ name + '.png');
            child = exec('curl -F file=@' + filepath  + ' -F channels=' + message.channel + ' -F token=' + process.env.token + ' https://slack.com/api/files.upload',
                function (err, stdout, stderr) {
                    if (err !== null) {
                        bot.botkit.log("Failed to add screenshot :(",err);
                    }
                }
            );
```

amacgregor:
I could never make the botkit method work so I ended up using a curl wrapper for the fileuploads that worked like a charm:

```
            var filepath = path.join(__dirname, '..', '/screenshots/'+ name + '.png');
            child = exec('curl -F file=@' + filepath  + ' -F channels=' + message.channel + ' -F token=' + process.env.token + ' https://slack.com/api/files.upload',
                function (err, stdout, stderr) {
                    if (err !== null) {
                        bot.botkit.log("Failed to add screenshot :(",err);
                    }
                }
            );
```

ZuzooVn:
Your method worked like a charm.
With https://www.npmjs.com/package/node-slack-upload, you can also add comment to your uploaded file.

RafaelCosman:
@amacgregor, I'm sorry that the Botkit method didn't work for you! I've labeled this as a bug.

peterswimm:
@amacgregor - did @mlallinder 's suggestion fix your issue? I am closing this for now, but please reopen if this is still an issue.

shawngiese:
Still having problems...
```
bot.api.files.upload({
[...]

PR #112, opened 2016-01-31
Added multi-part file upload to files.upload

olirogers:
Slight refactor of lib/Slack_web_api.js to include the ability to post data to Slack API as a multi-part upload to allow binary files to be updated.

The use is as follows:

```js
    bot.api.files.upload({
        file: fs.createReadStream('/tmp/dramacat.gif'),
        channels: 'C024BE91L,#general'
    }, function (error, done) {
        console.log(JSON.stringify(error));
        console.log(JSON.stringify(done));
    });
```

Is the same as (from [slack web api examples](https://api.slack.com/methods/files.upload#Examples))

```shell
curl -F file=@dramacat.gif -F channels=C024BE91L,#general -F token=xxxx-xxxxxxxxx-xxxx https://slack.com/api/files.upload
```

This fixes issue #29.

PR #391, opened 2016-08-29
Support `files.upload` API call.

sundeepgupta:
Fixes this old, closed issue: https://github.com/howdyai/botkit/issues/29
And is a follow up to this old PR: https://github.com/howdyai/botkit/pull/112

Added some robustness around posting to the Slack API by also accepting mutlipart form data, which is what the `files.upload` method requires. The reason why I didn't modify `callAPIWithoutToken`'s API  is because `files.upload` requires a user token.

### DIFF
--- a/lib/Slack_web_api.js
+++ b/lib/Slack_web_api.js
@@ -103,13 +103,17 @@ module.exports = function(bot, config) {
      * @param {Object} data The data to pass to the API call
      * @param {function} cb A NodeJS-style callback
      */
-    slack_api.callAPI = function(command, data, cb) {
-        if (!data.token) {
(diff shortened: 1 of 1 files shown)


sundeepgupta:
Any feedback/questions welcome, would love to see this merged in.

sundeepgupta:
:nudge:

benbrown:
I have added support for this in a slightly different style!

files.upload will now use multi-part uploads if the `file` parameter is passed, and normal form post if the `content` parameter is passed.

sundeepgupta:
Great, I like your approach more. Thanks!

PR #392, opened 2016-08-29
Facebook sender_action support ("is typing" indicator)

luisfarzati:
Implements `sender_action` parameter from the API as in: https://developers.facebook.com/docs/messenger-platform/send-api-reference#payload

Also adds convenience method `replyWithTyping`, following same convention than in Slackbot.

```javascript
 +bot.startTyping(message, function () {
 +  // do something here, the "is typing" animation is visible
 +});
 +
 +bot.stopTyping(message, function () {
 +  // do something here, the "is typing" animation is not visible
 +});
 +
 +bot.replyWithTyping(message, 'Hello there, my friend!');
 +```

### DIFF
--- a/lib/Facebook.js
+++ b/lib/Facebook.js
@@ -117,7 +117,7 @@ function Facebookbot(configuration) {
 
             var typingLength = Math.min(Math.floor(text.length / (avgCPM / 60)) * 1000, 5000)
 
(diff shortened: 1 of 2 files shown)


benbrown:
This is getting merged into 0.4! Thank you!

PR #397, opened 2016-08-30
Added delay capability to message.

tkornblit:
This will solve out of order messages for FB.

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -308,7 +308,7 @@ function Botkit(configuration) {
 
                 res[key] = {
                     question: this.responses[key].length ?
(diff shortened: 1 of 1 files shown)


Raniazy:
+1

anonrig:
Can you solve the conflicts in `lib/CoreBot.js`?

benbrown:
@tkornblit Can you explain to me a bit more why this is necessary?

There is already a facility for specifying a delay for each message that is relative to when the _previous message_ was sent.   This new code looks like it conflicts with that functionality...

benbrown:
We've solved this issue by adding support for message delivery confirmation in the latest version.

PR #398, opened 2016-08-30
Add Facebook read receipt event

etown:
Add support for read receipt events which were introduced in platform version 1.1. They work exactly like delivery receipts.


benbrown:
This was duplicated in another PR. Thank you!!

PR #404, opened 2016-09-02
Optionall allow non-numeric ports

### LINKED ISSUE #150: setupWebserver not working on Azure Website
slack_botkit.setupWebserver

process.env.PORT is not a number in an Azure Website, hence this check fails:
         if (isNaN(port)) {
             throw new Error('Specified port is not a valid number');
         }

Is there a way we can fix this?

stevegreatrex:
Added optional support for non-numeric ports for webserver so you can
run your slackbot on an Azure webapp (which uses a `\.\pipe\...` port
format)

Fixes #150

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -274,7 +274,8 @@ function Slackbot(configuration) {
         if (!port) {
             throw new Error('Cannot start webserver without a port');
         }
-        if (isNaN(port)) {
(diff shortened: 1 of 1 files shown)


stevegreatrex:
Hi @anonrig- was there something wrong with this PR? The use case is legitimate, I think

anonrig:
There wasn't a problem but there was a more cleaner approach in previous PR. Removing this check in overall is a much more cleaner approach. Therefore, I closed this PR. Sorry for the inconvenience and your contribution!

stevegreatrex:
No worries. As long as the support is there then I'm happy

PR #408, opened 2016-09-08
Also keep app access_token

PanMan:
Not only save bot token but also app access token as some calls need it and it can have different permissions


PanMan:
Can we do this? Or should I keep a fork? Any good reason not to do this? I need the access token to get individual messages.

benbrown:
Thanks, @PanMan! sorry for the delay.

PR #412, opened 2016-09-12
Adding SQLite storage support. Adding named pipe support.

MiroValchev:
1. I've added SQLite storage support. Simple JSON file support can't handle more than 2000 files. SQLite is simple and powerful. Should be available by default and not as a custom storage.
2. Azure and others are using named pipes and not port numbers. For that reason I've added a configuration that allows the isNaN(port) check to be skipped.


anonrig:
This should be done on a different repository. We're not currently accepting any new pull requests regarding storage support on Botkit.

MiroValchev:
@anonrig What about the named pipe change? It's a small change that would benefit many developers? I can create another pull request just with it, if you say you are gonna introduce it.
It's just this line:
https://github.com/MiroValchev/botkit/blob/36cb57d57cd8a3a44617dce7e5b63ae2ab3dbb58/lib/SlackBot.js#L277

anonrig:
Instead of a new variable, removing NaN is the appropriate solution for this. Because port is not always a number therefore NaN(Not a number) check is invalid. If you can open a new pull request to check only if port number exists or not, it will be enough for our current needs. @MiroValchev

PR #427, opened 2016-09-27
App ID & Password should be passed to spawn() 

Stevenic:
I was passing the App ID & Password to the controllers constructor and should have been passing it into spawn().


Stevenic:
This is just an error in the sample & docs so no rush.

Issue #369, opened 2016-08-14
webhooks failing on facebook messenger heroku hosted app


Now2407:
hello, my webhooks seem to be failing. facebook is sending me alerts saying my webhooks are failing when using this framework. i am hosting node js on heroku. could someone point me in the right direction as far as getting it working? thanks!

<img width="779" alt="screen shot 2016-08-13 at 10 09 35 pm" src="https://cloud.githubusercontent.com/assets/6015402/17647403/b0bce930-61a2-11e6-843d-543a935ee7a2.png">

yannderyckedotcom:
the is a problem with your app on heroku, is crashed.
you have to restart dynos and solve your erros
fb try to reach your app but can t

peterswimm:
Discussion continued in https://github.com/howdyai/botkit/issues/435

PR #435, opened 2016-10-09
Remove integer port requirement for Facebook

dwcares:
PaaS hosting offerings like Heroku and Azure App Service do not have a standard numerical port, as it's randomly generated when the app is re-started. 

The code currently has a check that ensures the port specified is a number. This will always fail and throw an exception on Heroku and Azure App Service, causing the app the crash. 

Example port from Azure App Service: "\\.\pipe\a59322f5-0005-4a3b-9242-24625191a3f2"

My fix removes the number check as ExpressJS, which is how the bots are served, has no requirement that ports are a number. 

Closes howdyai/botkit#369

### DIFF
--- a/lib/Facebook.js
+++ b/lib/Facebook.js
@@ -285,9 +285,6 @@ function Facebookbot(configuration) {
         if (!port) {
             throw new Error('Cannot start webserver without a port');
         }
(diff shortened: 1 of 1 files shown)


anonrig:
Nice catch!

PR #440, opened 2016-10-12
Don't use JSON.parse/stringify to clone.

### LINKED ISSUE #439: Conversation.cloneMessage throws away `Symbol` and `Date` objects
In working with the botkit I was trying to pass some custom middleware some `Symbol`s on the message object. By the time the middleware gets them, the `Symbol`s have all been `null`ed out.

Turns out that's because the `cloneMessage` method on Conversation uses `JSON.parse` and `JSON.stringify` to deep clone the message object. This method is great and really fast, except it discards values that aren't JSON-compatible, like `Date`s and `Symbol`s and `Function`s. Not cool, man, not cool.

drhayes:
Fixes #439.

Uses the `clone` npm package to provide a deep clone. This _probably_ has performance implications since `JSON.parse(JSON.stringify(thing))` is super fast, but I don't think the original implementation meant to throw away `Date` and `Symbol` objects.

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -8,6 +8,7 @@ var simple_storage = require(__dirname + '/storage/simple_storage.js');
 var ConsoleLogger = require(__dirname + '/console_logger.js');
 var LogLevels = ConsoleLogger.LogLevels;
 var ware = require('ware');
+var clone = require('clone');
 
 var studio = require('./Studio.js');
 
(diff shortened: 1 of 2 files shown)


ianp:
You could call it as `clone(message, true)` to get better performance, because stringily/parse would fail with circular references in any case.

Other than that, this looks good to me to merge.

benbrown:
Thanks @drhayes @ianp!

PR #441, opened 2016-10-13
Slack's `chat.update` should work with attachments

sundeepgupta:
When attachments are sent to these API methods, they need to be handled in the same way. `postMessage` was already doing it, so had `update` do the same.


sundeepgupta:
@peterswimm i should have been more clear, this is really a bug fix.

benbrown:
Thanks, @sundeepgupta

PR #446, opened 2016-10-16
Fix `bot.replyPrivateDelayed`. https://github.com/howdyai/botkit/issu…

sundeepgupta:
…es/444

I discovered this when trying to ephemerally respond to an interactive button click. We need to set `replace_original` to `false` in order for our reply to be ephemeral. Otherwise, it will reply publicly. Furrther, when responding to an ephemeral message's button click, currently we cannot update it. Thus, introduce new method to provide the option to update or not. The existing one could be depcrecated.

### DIFF
--- a/lib/Slackbot_worker.js
+++ b/lib/Slackbot_worker.js
@@ -494,6 +494,43 @@ module.exports = function(botkit, config) {
             msg.channel = src.channel;
 
             msg.response_type = 'ephemeral';
+            msg.replace_original = false;
+
+            var requestOptions = {
+                uri: src.response_url,
+                method: 'POST',
+                json: msg
+            };
+            request(requestOptions, function(err, resp, body) {
+                /**
+                 * Do something?
+                 */
+                if (err) {
+                    botkit.log.error('Error sending slash command response:', err);
+                    cb && cb(err);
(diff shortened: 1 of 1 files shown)


sundeepgupta:
Running into this again today. If we want to update an ephemeral interactive message, we need to be able to set `msg.replace_original = true`. I see three options to fix this...

1. Add API with something like `replyPrivateDelayedReplace`.
2. Change existing API by requiring a `replace_original` argument for `replyPrivateDelayed`. We could make this non-breaking by checking the `arguments`. 
3. Deprecate `replyPrivateDelayed` in favour of `replyPrivateDelayed2` (can't think of a name right now :))

I imagine the same will need to be done for `replyPublic`, `replyPublicDelayed`, and `replyPrivate` as well.

benbrown:
Hey @sundeepgupta  - I am not clear why we need a whole new function for this.

You can pass in the replace_original field as part of the parameter to replyPrivateDelayed...

```
bot.replyPrivateDelayed(incoming_message, {text: 'foo', replace_original: false});
```

sundeepgupta:
@benbrown not sure why i didn't see that.. my bad!

PR #448, opened 2016-10-17
Add a type field to Facebook postbacks

### LINKED ISSUE #447: No way to determine message type in message_received handler
For example, when a Facebook bot receives a post back it will trigger both the `facebook_postback` and `message_received` events, while this is good for conversational usage it means that messages will be processed twice in some cases. It would be good if there were an easy way to tell if a message had been generated from a post back in the message_received handler.

I'd like to add a `type` field to the message which would be set to `facebook_postback` (similar to how the Slack bot adds a `type` field to some messages).

(Botkit versions 0.4.0 and HEAD)

ianp:
Closes #447.


ianp:
Superseded by #469 .

benbrown:
Thanks for your work!

PR #454, opened 2016-10-18
Uppercase env vars

### LINKED ISSUE #437: Environment variables should be UPPERCASE
Specifically, the use of "port" for the environment variables in the examples that contain lines such as this one:
`if (!process.env.clientId || !process.env.clientSecret || !process.env.port) {`
Please change "port" to the more commonly used "PORT" variable.

Dinoshauer:
Heya! This fixes #437 -- I agree that they should be uppercase as this is what's normal for env vars :) Uppercasing and snakecasing them also makes sure they all look the same (twilio's env vars was already upper- and snakecased)

### DIFF
--- a/botframework_bot.js
+++ b/botframework_bot.js
@@ -23,7 +23,7 @@ This bot demonstrates many of the core features of Botkit:
 
   Run your bot from the command line:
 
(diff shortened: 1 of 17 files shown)


benbrown:
This is a good idea, but this is also a breaking change that will need to be carefully communicated when we merge this.

PR #459, opened 2016-10-24
Slack Events api

### LINKED ISSUE #438: Support for the Slack Events API
Botkit currently supports the RTM API in Slack which is based on WebSockets. Because of this we cannot use BotKit with AWS Lambda. Consider adding support for the  Event API of Slack.

jonchurch:
Receive Events API events at the webhook setup by `createWebhookEndpoints()`

EDIT: `startTicking()` is now called for you when wehbhook endpoints are created 
<s>Use `controller.startTicking()` when running bot to enable conversations</s>

Would like feedback on Bot Identity management without RTM. I'm saving bot identity to team storage when a user completes oauth flow, and looking it up from storage each time an Events API request is received.

closes #438


jonchurch:
Added a config option `eventsApi: true` to:
* Tell botkit to send messages via web API instead of RTM
* Ignore messages triggered by RTM, if the RTM is connected to (for purposes of presence) while using Events API

jonchurch:
Closing in favor of #480

PR #460, opened 2016-10-25
Added timestamp property to sent messages.

cwoolfRuxit:
Since no convo.askAndUpdate method exists, I am using the convo.ask method and updating the previously sent message using the Web API update method. It appears that messages from users contain a timestamp, but messages from bots don't. I added the ts (timestamp) property to messages sent and stored in the sent array. Please consider adding the timestamp to sent messages so they can be easily updated.

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -484,6 +484,7 @@ function Botkit(configuration) {
 
         this.tick = function() {
             var now = new Date();
+            var self = this;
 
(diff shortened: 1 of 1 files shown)


benbrown:
It is important that messages in conversations be able to be updated in this way.

However, this change will only work for Slack bots, and will break other types that do not return these fields.

We'll have to figure out another way to do this!

PR #465, opened 2016-10-31
Update slackbutton_bot.js example

### LINKED ISSUE #458: slackbutton_bot.js does not include redirectUri in config
I'm using botkit v 0.4.0 from NPM. 

https://github.com/howdyai/botkit/blob/master/examples/slackbutton_bot.js#L41

I just got a slackButton bot working from the slackbutton_bot.js starter code. When I was initially trying to get my bot to authenticate via the "login" url, I was getting an error "missing clientId in request". The issue seemed to resolve itself when I passed the redirectUri in as an environment variable. 

I think this should be added to the slackbutton_bot.js at the line highlighted in the snippet above. I can submit a fix once someone verifies that this is an actual issue.

jonchurch:
Adds forgotten line in example so it works as expected

botkit needs redirectUri in config for slackbutton oauth
Pass it into the example bot, check for redirectUri when run

closes #458

### DIFF
--- a/examples/slackbutton_bot.js
+++ b/examples/slackbutton_bot.js
@@ -27,8 +27,8 @@ This is a sample Slack Button application that adds a bot to one or many slack t
 /* Uses the slack button feature to offer a real time bot to multiple teams */
 var Botkit = require('../lib/Botkit.js');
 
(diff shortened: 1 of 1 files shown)


benbrown:
Thanks again, @jonchurch

PR #466, opened 2016-10-31
Update readme addQuestion() capture options docs

### LINKED ISSUE #456: Optional Parameter in addQuestion() requires a blank object to be passed to it
The addQuestion() method in conversations has an optional third parameter for capture_options and a required 4th parameter for thread_name. But it is required that a blank object is passed as the thirs parameter even if we dont want any capture_options or otherwise it takes the 4th parameter as the third and gives weird functionality.

jonchurch:
Adds clarification to addQuestion capture option docs. Must pass empty object if not using capture options

closes #456


benbrown:
Thanks @jonchurch

PR #469, opened 2016-11-01
Add a type field to message to distinguish between 'message_received' and 'postback'

agamrafaeli:
Apropo to issue #447

### DIFF
--- a/lib/Facebook.js
+++ b/lib/Facebook.js
@@ -220,7 +220,7 @@ function Facebookbot(configuration) {
                                 user: facebook_message.sender.id,
                                 channel: facebook_message.sender.id,
                                 timestamp: facebook_message.timestamp,
-                                type: "postback",
+                                type: "facebook_postback",
                             };
 
                             facebook_botkit.receiveMessage(bot, message);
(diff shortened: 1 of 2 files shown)


ianp:
See also PR #448 although the two are almost identical!

I think that I like my approach of reusing `facebook_postback` vs. `postback` for that type, but I also like that you've added a type field to non-postback messages, saves a null-check in client code.

agamrafaeli:
I agree re-naming convention. Personally I have an aversion to fields that require a null check. 

From what I can tell receiveMessage() is only called twice - on `message_received` and on `facebook_postback`. Thus adding type on both of them resolves the issue.

agamrafaeli:
This may be over-engineering at this point, but maybe they should both be string constants?

ianp:
I like to use string constants for stuff like this in my own projects, but it's not a common approach in JS land generally.

agamrafaeli:
So are we in agreement on the solution here? Do we still need two PRs?

ianp:
No, I've closed the other one, let's get this one merged 😀

agamrafaeli:
Sounds good to me. Got any info on the frequency of merges here?

ianp:
Looking at the number of open PRs and issues, not high! But I'll mention it in the Slack channel.

ianp:
In the doc changes you have “`type` will be set to `postback`” when it should say `facebook_postback`.

BTW talking with the Howdy guys now about getting some open PRs merged.

agamrafaeli:
@ianp changed. Good catch.

ianp:
This looks good to merge now.

benbrown:
Thank you guys! Sorry it took longer than it should have. Merged!!

Issue #463, opened 2016-10-27
RTM abnormal close 1006 - Messages not sent


rickul:
Hi!
We have a problem with the RTM. When starting some conversations, the RTM connections is (always) lost with a 1006 abnormal close. This only happens in one specific channel, in others this never happens.
When it happens, RTM is reconnected automatically, but the messages are not being sent. I think, they should be sent after a reconnect. A comment in bot.send() in Slackbot_worker.js says:
/**
- The RTM failed and for some reason it didn't get caught
- elsewhere. This happens sometimes when the rtm has closed but
- We are sending messages anyways.
- Bot probably needs to reconnect!
  */

rickul:
**Feature/Workaround Request**
Due to this bug we cannot trust in the RTM. But botkit conversations use both, RTM if possible and API calls if necessary.

Could you add something like a "forceAPICall" option to botkit to never use RTM:

```
var controller = Botkit.slackbot({
  ...
  forceAPICalls: true,
  ...
})
```

peterswimm:
Issue solved by #474

PR #474, opened 2016-11-04
force the bot sending messages through API; this commit closes issue …

rickul:
…#463


benbrown:
@rickul -- the purpose of this is to use the Web API to send messages instead of the RTM. Is that correct?

I have been considering making this the default method to send messages, as the web api supports a lot more formatting features.

rickul:
@benbrown Correct, I want to avoid sending messages through rtm and use the web API instead.

Would be great if you make it the default to use the Web API!

benbrown:
@rickul Sending via Web API is now the default!

I will be publishing this change to NPM later today.

PR #480, opened 2016-11-08
Slack Events api

### LINKED ISSUE #438: Support for the Slack Events API

jonchurch:
Had to resubmit this because I somehow got the other branch into a non-repairable headless state that github support advised I abandon ¯\_(ツ)_/¯ 

Receive Events API events at the webhook setup by `createWebhookEndpoints()`

EDIT: `startTicking()` is now called for you when wehbhook endpoints are created 
<s>Use `controller.startTicking()` when running bot to enable conversations</s>

Would like feedback on Bot Identity management without RTM. I'm saving bot identity to team storage when a user completes oauth flow, and looking it up from storage each time an Events API request is received.

closes #438


sundeepgupta:
Read the changelog earlier and the latest 4.1 release already support the Events API - no?

jonchurch:
Nope! It was mostly small fixes in 4.1
Biggest slack change was defaulting to Web API for sending messages, which is a separate issue from receiving events from Events API at a webhook.

Any feedback @sundeepgupta?
I still need to better explain subscribing to events in the docs. For example, a file shared into a channel the bot is in will generate a message event (assuming the bot is subscribed to messaging events) that has a subtype filed with a value of `file_share`. This is representative of the message slack sends into the channel. Botkit will trigger any message with a subtype as an event of that subtype. So now we can listen to `file_share` events. However, there is a separate permission in the Events API that is fired whenever a file is shared anywhere on a team. That will fire a `file_shared`(notice the d) event from the Events API, and botkit will fire `file_shared` when receiving this event. However the `file_shared` event is generated under a different permission, and contains different information than `file_share`(which at this point has much more data about that file than `file_shared`). 

I left it this way because that's how Botkit did it before, but this would likely be very confusing unless explained well in the docs.

sundeepgupta:
@jonchurch I've yet to learn about Events API and haven't looked at your PR, sorry. Thought I'd just check to make sure this wasn't duplicated effort but I obviously mis-understood the changelog.

lib/SlackBot.js
```diff
@@ -118,6 +118,73 @@ function Slackbot(configuration) {
...
+                }
+                // let's normalize some of these fields to match the rtm message format
+                message.team = req.body.team_id;
+                message.events_api = true;
+                message.authed_users = req.body.authed_users;
+                slack_botkit.findTeamById(message.team, function(err, team) {
+                    if (err || !team) {
+                        slack_botkit.log.error('Received Events API message, but could not load team');
+                        if (err) {
+                            slack_botkit.log.error(err);
```
colestrode: if you return here, then you save yourself the if and keeps the indentation flatter.

lib/SlackBot.js
```diff
@@ -118,6 +118,73 @@ function Slackbot(configuration) {
...
+                        // Identify the bot from either team storage or identifyBot()
+                        bot.team_info = team;
+                        bot.identity = team.bot.identity;
+
+                        if (team.bot.identity == undefined || team.bot.identity.id == null) {
+                            // API call to slack to identify bot user
+                            bot.identifyBot(function(err, identity) {
+                                bot.identity = identity;
+
+                                if (req.body.event.type === 'message') {
```
colestrode: this logic is duplicated, could we make this a function?

lib/SlackBot.js
```diff
@@ -118,6 +118,73 @@ function Slackbot(configuration) {
...
+                        }
+                    } else {
+                        var bot = slack_botkit.spawn(team);
+                        // Identify the bot from either team storage or identifyBot()
+                        bot.team_info = team;
+                        bot.identity = team.bot.identity;
+
+                        if (team.bot.identity == undefined || team.bot.identity.id == null) {
+                            // API call to slack to identify bot user
+                            bot.identifyBot(function(err, identity) {
```
colestrode: what happens if there is an error identifying the bot?

examples/slackbutton_bot.js
```diff
@@ -27,8 +27,8 @@ This is a sample Slack Button application that adds a bot to one or many slack t
 /* Uses the slack button feature to offer a real time bot to multiple teams */
 var Botkit = require('../lib/Botkit.js');
 
-if (!process.env.clientId || !process.env.clientSecret || !process.env.port) {
-  console.log('Error: Specify clientId clientSecret and port in environment');
+if (!process.env.clientId || !process.env.clientSecret || !process.env.port || !process.env.redirectUri) {
```
jonchurch: This was actually part of a PR from #458, which I realize now was not necessary.
You are correct, redirectUri is not required here, I misunderstood their issue. 
Removing

benbrown:
This got merged in to #500 and is now released! Thanks @jonchurch and @colestrode and @sundeepgupta and @peterswimm for your attention to this!!

Issue #492, opened 2016-11-13
Default hostname should be 0.0.0.0 not 127.0.0.1


EliKrumholz:
The current default hostname is set to 127.0.0.1, which causes errors when deploying to heroku, unless it is overridden with 0.0.0.0. 

The node http module, which express inherits from, uses 0.0.0.0 as the default (as described here: https://nodejs.org/api/http.html#http_server_listen_port_hostname_backlog_callback)

> If the hostname is omitted, the server will accept connections on any IPv6 address (::) when IPv6 is available, or any IPv4 address (0.0.0.0) otherwise.

danielckv:
+1

aj-may:
+1 same issue with docker

peterswimm:
PR above has been merged, pending next update.

PR #493, opened 2016-11-13
Set 0.0.0.0 to default hostname

EliKrumholz:
See issue #492. Using a default hostname of 127.0.0.1 overrides the http module's expected default hostname as described here: https://nodejs.org/api/http.html#http_server_listen_port_hostname_backlog_callback. This can cause deployment errors on services such as heroku.

> If the hostname is omitted, the server will accept connections on any IPv6 address (::) when IPv6 is available, or any IPv4 address (0.0.0.0) otherwise.

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -1036,10 +1036,12 @@ function Botkit(configuration) {
 
     botkit.config = configuration;
 
-    // default the application to listen to the localhost IP address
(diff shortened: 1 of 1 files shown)


aj-may:
👍

benbrown:
@EliKrumholz Thank you for catching this!

PR #505, opened 2016-11-22
Make stale connection detection configurable

mbrevoort:
A Beep Boop user ran into an issue when on connection and periodic points in time, they were doing A LOT of work that was maxing out CPU so much that it caused node.js's event loop to queue excessively. RTM connections were healthy and returning `pong` responses but the node process wasn't processing the `pong` response in time because the event queue was so backed up. This caused a lot of thrashing of connects/discconects/connects within their process:

thrash

screenshot 2016-11-16 15 06 37

Ultimately they need to better manage/throttle the work being done inside of the process, but the mitigation to avoid the connection thrashing was to increase the amount of time considered to consider a connection stale.

An overview of the changes:

- add `stale_connection_timeout` to make stale connection detection configurable. This is the number of milliseconds to wait for a `pong` response to a `ping`
- replace the use of an `interval` with a `timeout` to ensure we don't clog the event loop under duress.
- add details to the Slack Readme

### DIFF
--- a/lib/Slackbot_worker.js
+++ b/lib/Slackbot_worker.js
@@ -21,7 +21,7 @@ module.exports = function(botkit, config) {
     // Set when destroy() is called - prevents a reconnect from completing
     // if it was fired off prior to destroy being called
     var destroyed = false;
-    var pingIntervalId = null;
+    var pingTimeoutId = null;
     var retryBackoff = null;
 
     // config.retry, can be Infinity too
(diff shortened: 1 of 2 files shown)


jonchurch:
Hey this looks good! It's a weird edge case for sure, but seems reasonable.

Any reason you took out the PING SENT and PONG RECEIVED debug messages?

PR #510, opened 2016-11-23
Adding redirectUri parameter to start bot command

musicaljoeker:
This node command fails without the redirectUri parameter being included.


jonchurch:
This was a mistake I introduced, because you don't *actually* need the redirectUri for the slack_button flow, it should not be necessary to run the file.

musicaljoeker:
When I tried running the starting command as instructed, it gave me an error saying that it needed the redirectUri parameter. The code that I updated was the only way for me to get the bot to launch.

jonchurch:
You are spot on that this is a bug, but the right fix is to make redirectUri optional and not required to run the file.

musicaljoeker:
Gotcha. That would just mean removing it from the first if-statement, correct?

jonchurch:
Correct, I submitted #526 to address this.

PR #514, opened 2016-11-26
Fix bot_channel_join and bot_group_join with Slack Events API

### LINKED ISSUE #347: bot_channel_join event not firing when bot is invited to a channel
I am listening to the `bot_channel_join` event as follows:

```
controller.on('bot_channel_join',
    function(bot, message) {
        console.log(message);

        // reply with excitement
        bot.reply(message, It\'s me!!');
});
```

However, when I invite the bot to a channel the event doesn't fire.

jonchurch:
Events API does not pass bot_id to channel_join and group_join events, which we were checking for in order to determine if it was the bot which joined the channel or group. 

This restores expected behavior

I added an exception for channel and group joins to filtering out events associated with our bot user, I think this is the only event we listen to with slack that has our bot's user id as the `user: id` so I think this should be fine for now.

closes #347

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -228,8 +228,8 @@ function Slackbot(configuration) {
                     id: team.bot.user_id,
                     name: team.bot.name
                 };
+                if (team.bot.user_id === req.body.event.user && req.body.event.subtype !== 'channel_join') {
 
-                if (team.bot.user_id === req.body.event.user) {
                     slack_botkit.debug('Got event from this bot user, ignoring it');
                     return;
                 }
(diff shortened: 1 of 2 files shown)


lib/SlackBot.js
```diff
@@ -228,8 +228,8 @@ function Slackbot(configuration) {
                     id: team.bot.user_id,
                     name: team.bot.name
                 };
+                if (team.bot.user_id === req.body.event.user && req.body.event.subtype !== 'channel_join' && req.body.event.subtype !== 'group_join') {
```
chrisberthe: Remove line.

JimTheMan:
I am so confused. The code is still checking for `team.bot.user_is`, and the `bot_channel_join` event is still not working in the botkit starter apps...

PR #515, opened 2016-11-26
Trigger Events from ConsoleBot

### LINKED ISSUE #347: bot_channel_join event not firing when bot is invited to a channel

jonchurch:
Add ability to trigger events from ConsoleBot. 
Will trigger event type that follows a space after `trigger:`. Must use colon to match, must be first word on the line. 

This does not actually fire a message with the correct associated data for this event (like channel info, message, etc) will just invoke any listeners attatched to that event name you have created. 

Just another tool for console bot ❤️ 🤖 
closes #347

### DIFF
--- a/console_bot.js
+++ b/console_bot.js
@@ -66,7 +66,7 @@ var controller = Botkit.consolebot({
 var bot = controller.spawn();
 
 controller.on('bot_channel_join', function(bot, message) {
-    console.log('listener fired!')
+    // console.log('listener fired!');
     bot.send({text: 'Here I am in a channel now!'})
 });
 
--- a/console_bot.js
+++ b/console_bot.js
@@ -65,6 +65,11 @@ var controller = Botkit.consolebot({
 
 var bot = controller.spawn();
 
+controller.on('bot_channel_join', function(bot, message) {
+    console.log('listener fired!')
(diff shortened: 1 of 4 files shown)


skw:
Nice! Thanks for fixing this @jonchurch :)

btw isn't this more of a bug than an enhancement?

jonchurch:
Triggering events from ConsoleBot is a feature I'd say, I think it was mainly intended to test conversations until now. Triggering events isn't really that useful, unless you are passing dummy data with the event. But at least you can test the logic for rapid prototyping 😉 

I had some unrelated changes that involved fixing bot_channel_join events under slack Web API from previous commits, undid them in a later commit. This deals just with adding event triggering capability to ConsoleBot. 👍

alecl:
This seems worth merging in. Any interest?

benbrown:
Dang, @jonchurch this is a good idea! However, I think it would be better implemented as part of the console_bot example instead of a core feature.

Thoughts?

PR #525, opened 2016-12-02
slackbutton_bot.js redirectUri Bug Fix

jonchurch:
Closes #510 

I introduced redirectUri as a required variable to run slackbutton_bot.js example. redirectUri is an optional parameter that is part of the slackbutton oauth flow, and as such should not be required to run the file.

This fixes that, and comments out redirectUri in the bot's config while pointing it out as an optional parameter.


jonchurch:
Resubmitting this instead of rebasing around my local changes.

PR #532, opened 2016-12-05
FIX RTM RETRY AND RECONNECT LOGIC

inn0v8:
Existing Botkit slackbot reconnect logic is failing. The retry attributes are passed along with the botkit object but slackbot_worker.js is looking for them in the bot object. Because these attributes do not exist the retry will always fail. There have been several issues opened against this problem and custom code has been written to try to manage the scenario with there is an RTM disconnect and the bot essentially dies for that team because it cannot reconnect.

This fix points to the botkit object in slack_worker.js which then allows the retry logic to work as planned. Since making this change the bot will automatically reconnect and use the reconnect attributes that are used when the controller is initiated.

### DIFF
--- a/lib/Slackbot_worker.js
+++ b/lib/Slackbot_worker.js
@@ -25,8 +25,8 @@ module.exports = function(botkit, config) {
     var retryBackoff = null;
 
     // config.retry, can be Infinity too
-    var retryEnabled = bot.config.retry ? true : false;
-    var maxRetry = isNaN(bot.config.retry) || bot.config.retry <= 0 ? 3 : bot.config.retry;
+    var retryEnabled = botkit.config.retry ? true : false;
+    var maxRetry = isNaN(botkit.config.retry) || botkit.config.retry <= 0 ? 3 : botkit.config.retry;
 
     /**
      * Set up API to send incoming webhook


jonchurch:
@benbrown This looks like the same change I have made in #514, looking for config vars on `botkit.config` instead of `bot.config`

The [docs](https://github.com/howdyai/botkit/blob/master/readme-slack.md#controllerspawn) specify passing `retry`when spawning a bot, and changing the logic in this way could be breaking for some folks. 

But I think most people have been trying to pass their retry options into the controller or `configureSlackApp({})` because the spawning logic for Slack apps out of the box is abstracted away into getting the config from the DB then connecting all existing bots. 

I'd propose changing convention so `bot.spawn()` only deals with identifying a bot, giving it either a token or a team config with bot.identity. Options like retry and send_via_rtm seem like global options for the controller to me, and that's where I naturally want to pass them into.

gunar:
@benbrown Looking forward to this.

benbrown:
I agree @jonchurch those type of options belong at the controller level.

However, for backwards compatibility and also to handle some multi-team bots, we should allow it to be passed in at the bot level too. 

I'll look int the code and see what needs to happen to make this mergable.

benbrown:
OK! I did this in a slightly different way, but should be good to go.

gunar:
@benbrown do you mean [here](https://github.com/howdyai/botkit/commit/2e73bcf17962897fd3b608773214f848045f06da)?

benbrown:
@gunar yes

PR #539, opened 2016-12-08
Ignore confirmation messages in RTM message receive

uzairfarooq:
Slack RTM api sends back a confirmation message for each message sent out (see **Handling responses** section at https://api.slack.com/rtm). This causes `controller.middleware.receive.use(function(bot, message, next) {})` method to fire for confirmation messages too.

For example, if bot sends a message **Hello uzair**, slack would send back a confirmation message with the same contents (i.e. **Hello Uzair**). In result the `controller.middleware.receive` will fired for the confirmation message **Hello uzair**


benbrown:
This is a fair idea, but right now, this would interfere with the message delivery confirmation code.

One way to approach this would be to fire some other type of event for the reply_to messages, and to use that new event for delivery confirmation.

PR #541, opened 2016-12-09
Highlighted "default" as part of the actions

abinashmohanty:
It's intuitive to highlight keywords `default` and `next_step` as part of the Actions as documented under convo threads.


benbrown:
@abinashmohanty though `default` is the name of the default conversation topic, next_step is not actually anything in this context. Not sure what you are trying to highlight...

PR #548, opened 2016-12-12
Add removeListener function for event handlers

kjdelisle:
A removeListener function that lets the API user provide the
function they're using to handle .hears and .on events in a way
that lets them remove that event without needing to dive into
the underlying mechanisms in CoreBot.


kjdelisle:
Essentially, I just aligned a 2nd array that contains the callbacks provided by users (referred to as listeners) with the events arrays so that we can remove from both arrays using the same index. This way, you can dynamically remove those events without needing to traverse those collections yourself.

kjdelisle:
Updated the PR to reorder the parameters and add a check in case the 3rd parameter is undefined.

PR #554, opened 2016-12-13
Add in-memory storage to simple_storage.js

ErinsMatthew:
Add the ability to leverage JFS's in memory storage option within simple_storage.js.


lib/storage/simple_storage.js
```diff
@@ -28,13 +28,23 @@ module.exports = function(config) {
 
     if (!config) {
         config = {
-            path: './',
+            path: './'
```
ErinsMatthew: Removed extraneous comma after last list item.

lib/storage/simple_storage.js
```diff
@@ -28,13 +28,23 @@ module.exports = function(config) {
...
         };
     }
 
-    var teams_db = new Store(config.path + '/teams', {saveId: 'id'});
-    var users_db = new Store(config.path + '/users', {saveId: 'id'});
-    var channels_db = new Store(config.path + '/channels', {saveId: 'id'});
+    config.path = config.path || './';
```
ErinsMatthew: Set default for config.path if not set. This was necessary in case someone failed to include config.path when they passed in config for the new in-memory option.

lib/storage/storage_test.js
```diff
@@ -31,7 +31,8 @@ var testStorageMethod = function(storageMethod) {
                 test.assert(data.foo === testObj0.foo);
             });
             storageMethod.get('shouldnt-be-here', function(err, data) {
-                test.assert(err.displayName === 'NotFound');
+                //handle JFS get issue for in-memory returning empty err object
```
ErinsMatthew: I will be submitting a PR for the JFS project as well.

PR #555, opened 2016-12-14
DDOS Vulnerability Fix - Secure Facebook Webhook

jonchurch:
**Proposed Changes:** add express middleware to webhook POST route to validate incoming requests. Validate the X-HUB signature sent in headers against your Facebook App Secret. Throw an error to abort parsing requests that failed validation. The error is handled by an error handler which responds with a safe error to unvalidated requests. Relies on node's crypto lib. 

Configurable by passing `validate_requests: true, app_secret: your_app_secret` when creating your controller object.

When in production, this prevents an attacker from forcing your application to process huge arrays of requests or impersonate FB/users. FB batches its messages, and so a malicious server could make requests stuffed full of data that your bot would then iterate over, potentially locking the process.

### DIFF
--- a/facebook_bot.js
+++ b/facebook_bot.js
@@ -76,6 +76,10 @@ if (!process.env.verify_token) {
     process.exit(1);
 }
 
+if (!process.env.app_secret) {
+    console.log('Error: Specify app_secret in environment');
+}
+
 var Botkit = require('./lib/Botkit.js');
 var os = require('os');
 var commandLineArgs = require('command-line-args');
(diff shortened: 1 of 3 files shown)


benbrown:
@jonchurch this looks great, would love to merge but for a few small tidbits:

1) instead of using the process.env object to get at the secret, can you make that a new key in the controller's config object? 

2) if in the example app you want to default it to true -- i think this is fine -- we should also make sure it has the appropriate value for its secret and error and exit the process if not.

jonchurch:
@benbrown Fixed things up! 

Facebook_bot.js requires `app_secret` now. 

In addition to your changes, I switched `fb_app_secret` to `app_secret` to match convention so far.

PR #566, opened 2016-12-22
Add custom api_host env variable

afilatov1989:
Allows users to change default "slack.com" api host to any custom host via env variable.
Can be useful for various api proxies needed by different services.

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -338,7 +338,8 @@ function Slackbot(configuration) {
     slack_botkit.getAuthorizeURL = function(team_id, redirect_params) {
 
         var scopes = slack_botkit.config.scopes;
-        var url = 'https://slack.com/oauth/authorize' + '?client_id=' +
+        var api_host = process.env.api_host ? process.env.api_host : 'slack.com';
+        var url = 'https://' + api_host + '/oauth/authorize' + '?client_id=' +
             slack_botkit.config.clientId + '&scope=' + scopes.join(',') + '&state=botkit';
 
         if (team_id)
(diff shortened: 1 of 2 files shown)


jonchurch:
Can you help me understand what use case you have in mind for this?

Slackbot relies on the default slack API, not sure when an average user would ever want to change that. What are you building? 😄

katsgeorgeek:
average user maybe not, but why not to have ability to change it when you need it? as I mentioned in previous PR my team needs it for testing

lib/SlackBot.js
```diff
@@ -338,7 +338,8 @@ function Slackbot(configuration) {
     slack_botkit.getAuthorizeURL = function(team_id, redirect_params) {
 
         var scopes = slack_botkit.config.scopes;
-        var url = 'https://slack.com/oauth/authorize' + '?client_id=' +
+        var api_host = process.env.api_host ? process.env.api_host : 'slack.com';
```
katsgeorgeek: I think it could be more readable `process.env.api_host || 'slack.com';`

lib/Slack_web_api.js
```diff
@@ -13,8 +13,9 @@ function noop() {}
  * @returns {Object} A callback-based Slack API interface.
  */
 module.exports = function(bot, config) {
+    var api_host = process.env.api_host ? process.env.api_host : 'slack.com';
```
katsgeorgeek: The same as I mentioned above

afilatov1989:
We are building a service for deploying bots to different platforms.

E.g. we have a Slack bot and want it to work on Skype.
So we provide a proxy api, which receives requests instead of a real Slack api, translates it to Skype format, and sends to Skype. Response is translated and transmitted the same way.

Link to the service: https://message.io/

lib/SlackBot.js
afilatov1989: I agree that it's more concise, but I think it's less readable, because the result is pretty unpredictable for js newbies. In many languages this statement will proceed boolean true, which will be converted to anything like 1, "true", etc.

colestrode:
If you are wanting to change the URL for testing, have a look at https://github.com/Skellington-Closet/slack-mock That will intercept your requests without needing to change the URL.

afilatov1989:
No it's not for testing. It will be used on production for switching bots from real slack api to the proxy, which I described before.

peterswimm:
@afilatov1989 It would probably be better to add message.io support as a intergration akin to how we implented twilio or facebook versus hacking slack specific .js files. This way you preserve slack functionality, and users can start with a message.io starter kit we can host in the project if that is how they want to start building their bot.

benbrown:
This is included in 0.5!

PR #568, opened 2016-12-23
Add support for Facebook location quick replies 

mvaragnat:


mvaragnat:
Adding default to text if type is not defined

PR #586, opened 2017-01-05
adds createPrivateConversation to Slack bots

### LINKED ISSUE #585: Add a createPrivateConversation() method for Slack bots
Slack bots can use `startPrivateConversation()` to initiate a conversation in a DM with a user. I'd like to use conversation threads to set up a complex DM conversation with a user. I suggest we add a `createPrivateConversation()` method analogous to `createConversation()`. `createPrivateConversation()` would be Slack only, just like `startPrivateConversation()`

colestrode:
This adds a `createPrivateConversation` method for Slack bots so that complex private conversations can be made with threads, etc.

Closes #585

### DIFF
--- a/lib/Slackbot_worker.js
+++ b/lib/Slackbot_worker.js
@@ -305,6 +305,16 @@ module.exports = function(botkit, config) {
         botkit.createConversation(this, message, cb);
     };
 
+    bot.createPrivateConversation = function(message, cb) {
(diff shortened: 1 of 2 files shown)


benbrown:
thanks @colestrode!

colestrode:
You're welcome! Thanks for reviewing/merging 😄

Issue #590, opened 2017-01-10
Cannot read property 'user_id' when dealing with Slack Apps.


nuutco:
When attempting to configure a private Slack App, I'm running into issues with `team.bot.user_id` being undefined? Is this a configuration issue with Slack Apps? The `team` object I get back in that area of code looks like this:

```
{
  "id": "UNIQID1",
  "createdBy": "UNIQID2",
  "url": "https://COMPANY.slack.com/",
  "name": "COMPANY"
}
```

I've pasted the error logs below.

```
TypeError: Cannot read property 'user_id' of undefined
    at /Users/knuut/slack-bot/node_modules/botkit/lib/SlackBot.js:177:33
    at Object.get (/Users/knuut/slack-bot/node_modules/botkit/lib/CoreBot.js:822:17)
    at Object.Slackbot.slack_botkit.findTeamById (/Users/knuut/slack-bot/node_modules/botkit/lib/SlackBot.js:300:36)
    at Object.Slackbot.slack_botkit.handleWebhookPayload (/Users/knuut/slack-bot/node_modules/botkit/lib/SlackBot.js:158:22)
    at /Users/knuut/slack-bot/node_modules/botkit/lib/SlackBot.js:137:26
    at Layer.handle [as handle_request] (/Users/knuut/slack-bot/node_modules/express/lib/router/layer.js:95:5)
    at next (/Users/knuut/slack-bot/node_modules/express/lib/router/route.js:131:13)
    at Route.dispatch (/Users/knuut/slack-bot/node_modules/express/lib/router/route.js:112:3)
    at Layer.handle [as handle_request] (/Users/knuut/slack-bot/node_modules/express/lib/router/layer.js:95:5)
    at /Users/knuut/slack-bot/node_modules/express/lib/router/index.js:277:22
```

I've also added the basic code used for testing.

```
const Botkit = require('botkit');

const controller = Botkit.slackbot({
  debug: true,
});

controller.configureSlackApp({
  clientId: process.env.CLIENT_ID,
  clientSecret: process.env.CLIENT_SECRET,
  scopes: ['incoming-webhook','team:read','users:read','channels:read','im:read','im:write','groups:read','emoji:read','chat:write:bot'],
});

controller.setupWebserver(process.env.PORT, (err, webserver) => {
  controller.createWebhookEndpoints(webserver);
  controller.createOauthEndpoints(webserver, (err2, req, res) => {
    if (err2) {
[...]

PR #591, opened 2017-01-10
Fix slackbutton_slashcommand example

leighman:
The example seems to need 'bot' in its scopes (since it uses replyX?). See https://github.com/howdyai/botkit/issues/590.

Based on another example I think (?) you need to use replyXDelayed after the first replyX. Without this change the second reply does not work and prints `Can't set headers after they are sent.` Error.


benbrown:
This was fixed by another pull request, I believe!

PR #597, opened 2017-01-15
Add clear event method to controller.

colus001:
### Why
- There's no way to clear events

### What I did
- Add `.clear` to remove existing events.

### To do (maybe me?)
- Clear event by callback

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -201,6 +201,17 @@ function Botkit(configuration) {
             return this;
         };
 
+        this.clear = function(event) {
+            botkit.debug('Clearing a handler for', event);
+            var events = event.split(/\,/g);
+            for (var e in events) {
+                if (this.events[events[e]]) {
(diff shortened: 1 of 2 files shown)


colus001:
I think it is necessary to change `.on(event, callback)` method to use some `id` so that it can be cleared by the reference.

colestrode:
The Node EventEmitter API provides an `removeListener` method, which takes an event type and a reference to the event callback, that signature might be a more intuitive API for Node developers. There is also a `removeAllListeners` which just takes an event type and removes all handlers: 

* https://nodejs.org/api/events.html#events_emitter_removelistener_eventname_listener
* https://nodejs.org/api/events.html#events_emitter_removealllisteners_eventname

colus001:
@colestrode That would be better. I'll change it soon. :D

jonchurch:
@colestrode any interest in picking this back up? I saw someone asking about this functionality recently

PR #607, opened 2017-01-19
adding timeout handler for custom handling for conversation timeout

manojkumar1412:
I have  added a handler for timeout of a conversation. It will give extra handling over timeout of a conversation. earlier flow has not been altered.


benbrown:
thanks for adding this @manojkumar1412!

This will make it into the next release of Botkit.

Issue #600, opened 2017-01-16
Problem listening for events related to the bot user


arnthorsnaer:
I'm currently trying to subscribe to Slack bot user's `im_open` event. However, as far as I can see, Botkit ignores events that involve the bot user - therefore the events are not triggered.

Am I missing something? Has Botkit not been updated to facilitate the bot user events?
It's easy enough for me to fork the code to fix this but I'd rather know the intention behind this and act accordingly.

https://github.com/howdyai/botkit/blob/master/lib/SlackBot.js#L246
```
if (bot.identity == undefined || bot.identity.id == null) {
    slack_botkit.log.error('Could not identify bot');
    return;
} else if (bot.identity.id === message.user && message.subtype !== 'channel_join' && message.subtype !== 'group_join') {
    //HERE IS WHERE BOT USER EVENTS GO TO DIE 
    slack_botkit.debug('Got event from this bot user, ignoring it');
    return;
} else {
    if (message.type === 'message') {
        slack_botkit.receiveMessage(bot, message);
    } else {
        slack_botkit.trigger(message.type, [bot, message]);
    }
}
```

benbrown:
Botkit ignores most events pertaining to the ot itself. 

If you've got a good reason for this event to pass through, submit a pull request with the necessary changes!

arnthorsnaer:
Yes I have. My use case is that the bot should greet users when they open a "direct message" dialog with the bot. The greeting contains instructions in available actions.

I'll submit a PR, thanks for the reply. I wonder if I should prefix these events with something so that existing users wont be caught off guard by these changes.... perhaps something to discuss on the PR.

PR #611, opened 2017-01-20
Add option to let bot user events pass through

arnthorsnaer:
Bot user events are currently blocked. I added the option to let them through (using boolean attribute receive_bot_user_events on the configuration object). 
https://github.com/howdyai/botkit/issues/600


arnthorsnaer:
I'm closing this pull request since it turns out the bot user events emitted are limited (example im_open is only emitted if the user associated with the bot is interacting with the bot). So since this won't work for our scenario, I doubt it will be helpful for others.

Issue #601, opened 2017-01-17
Responds with 401 unauthorised on /login when using verification token


osh123:
Setting up the controller with a Slack verification token, like this:

`    controller.createWebhookEndpoints(controller.webserver, ["xxxx"]);
`
Makes botkit responds with 401 for requests to /login.

havardthom:
I encountered this issue today, was the fix applied?

peterswimm:
@nishant-chaturvedi ?

nishant-chaturvedi:
Noticed that, this issue is still present, made the fix and opened the PR https://github.com/howdyai/botkit/pull/1203

PR #620, opened 2017-01-24
Fix slack auth issue, caused by auth middleware being applied globally

nishant-chaturvedi:
It's a fix for issue #601, this is specific to slack, in case of setting up webhook endpoint with verififcationToken enabled, slack auth middleware was getting applied globally for all the endpoints which was causing all requests getting refused with 401 response code.

Ideally this middleware should only be enabled for webhook endpoint, which is called by slack and verification token would be provided

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ examples/db_team_bot/
 .DS_Store
 */.DS_Store
 .env
+.vscode/settings.json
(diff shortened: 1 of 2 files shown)


nishant-chaturvedi:
closing this for now, will make another pull request after fixing it in the latest botkit version

PR #624, opened 2017-01-24
ambiguity in signature of delete functions in simple storage

znat:
delete functions accept user/team/channel_id, but then treats arguments
as objects and try to get an id from them.


colestrode:
This seems like a breaking change to the API, it would be safer to change the parameter name to reflect the method accepts and object.

benbrown:
An identical PR just got merged. Thanks folks!

PR #639, opened 2017-02-03
Add support for channels.replies method

miklosaubert:
Please add support for [channels.replies](https://api.slack.com/methods/channels.replies). As I understand it, this method can work with the default behaviour from `slack_api.callAPI`, at least it works fine for my app.


jonchurch:
👍 good catch!

benbrown:
thanks @miklosaubert

PR #640, opened 2017-02-03
Fix a couple typos for Slack

masterful:
A, uh, friend of mine, copied the url over when configuring and it unfortunately had the misspelled version of `receive`. Whoops.


peterswimm:
Good catch!

PR #644, opened 2017-02-06
add facebook page_id to incoming message as page property

smalltalk-josh:


smalltalk-josh:
For the rare case when the page_id value is needed. The specific case is when integrating with tools that will take over the conversation with the user. In order to pass control to a human agent, the 3rd party tool is asking for page_id (and other data which is already in the message).

devhaoy:
Cool, thanks for the good job! Do you know of any easy way to add a working Human In The Loop functionalities to Botkit?

benbrown:
Thanks @smalltalk-josh

PR #657, opened 2017-02-15
Store state parameter from oauth against team

davidwneary:
I display the 'Add to slack' button on my website and submit the logged in user's id using the `state` parameter when clicked. However, it appears that parameter is currently not accessible in botkit. This fix works locally in my certain case, but as I'm new to botkit there may be issues with this implementation that I'm unaware of so I'd appreciate any feedback.

Thanks

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -526,6 +526,8 @@ function Slackbot(configuration) {
                                         name: identity.team,
                                     };
                                 }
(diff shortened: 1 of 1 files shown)


alecl:
This seems worthwhile for allowing more data to be passed through on bot add.

benbrown:
thank you!

PR #659, opened 2017-02-15
conv.setTimeout(timeout) method documented.

tridib:
I had to do google to figure out out to set timeout for a conversation. I see frustration in issue section not being documenting this. We have such useful method. So, I documented in front page.


benbrown:
Thank you @tridib!

PR #662, opened 2017-02-16
Properly handling Facebook attachements (Iimage + audio and video)

ouadie-lahdioui:
Hello guys,

I want to share with you how i handle an image, a video or an audio received by the bot.

Facebook support multiple types multimedia and templates attachments (image, audio, video), when the bot user is supposed to type a simple text but they send an image instead, Facebook provide an  attachements key in JSON sent by it and let text one empty.

what do you think ?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ examples/db_team_bot/
 .DS_Store
 */.DS_Store
 .env
+.idea/
--- a/facebook_bot.js
+++ b/facebook_bot.js
@@ -394,6 +394,9 @@ controller.on('message_received', function(bot, message) {
     return false;
 });
 
+controller.on('facebook_image', function(bot, message) {
+    bot.reply(message, 'Very nice image !');
+});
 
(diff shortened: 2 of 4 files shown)


jonchurch:
Looks good!

I think users can only send one attachment per message currently, and as long as thats true then triggering one per present attachment is fine. 

If fb introduced group chats with multi bots, we wouldnt want to trigger one per attachment.

devhaoy:
Works perfectly but it seems to preempt any image received inside a convo. Some might prefer to handle attachment in their app.js to avoid this.

benbrown:
I have something like this built into the Facebook starter kit. I share @devhaoy's concern that this could impact the ability to receive media inside conversations...

Should this be built into Botkit, or be something noted in the docs and included in the example apps?

ouadie-lahdioui:
I'll take this into consideration on my Botkit fork, @benbrown i just whant to share with you how i do that and get a feedback !

PR #689, opened 2017-03-03
Add Messenger code API

ouadie-lahdioui:
Hello,

Messenger Codes can be scanned in Messenger to instantly link the user to your bot, no typing needed. They're great for sticking on fliers, ads, or anywhere in the real world where you want people to try your bot.

This PR add the ability to programmatically get your bot's Messenger Code.

Enjoy,

### DIFF
--- a/lib/Facebook.js
+++ b/lib/Facebook.js
@@ -517,6 +517,40 @@ function Facebookbot(configuration) {
                         }
                     });
             }
+        },
+        get_messenger_code: function(image_size, cb) {
+            var message = {
+                'type': 'standard',
+                'image_size': image_size || 1000
+            };
+            request.post('https://graph.facebook.com/v2.6/me/messenger_codes?access_token=' + configuration.access_token,
+                {form: message},
+                function(err, res, body) {
+                    if (err) {
+                        facebook_botkit.log('Could not configure get messenger code');
+                        cb(err);
+                    } else {
+
+                        var results = null;
+                        try {
+                            results = JSON.parse(body);
+                        } catch (err) {
+                            facebook_botkit.log('ERROR in messenger code API call: Could not parse JSON', err, body);
+                            cb(err);
+                        }
+
+                        if (results) {
+                            if (results.error) {
+                                facebook_botkit.log('ERROR in messenger code API call: ', results.error.message);
+                                cb(results.error);
+                            } else {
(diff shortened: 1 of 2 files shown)


jonchurch:
👏👏👏
This is awesome! Didn't even realize this api call existed, is it new?

I'm maintaining a Botkit fork right now that supports many pages at once, and so I've been slowly removing all bot identity config (like access_token and studio token) from the controller config.
So I'd rewrite this immediately to be scoped to the bot, I might be an edge case, not sure how often people are using bots that are unconfigured/generic these days. 

Personally I would add this to the bot object, and get at the access_token like so: 
```javascript

request.post('https://graph.facebook.com/v2.6/me/messenger_codes?access_token=' + configuration.access_token || this.config.access_token, {
             form: message
         },
```

What do y'all think? I think this is a great feature regardless of where it lives.

ouadie-lahdioui:
Yoo @jonchurch,

yes it's new, Facebook maked some changes last thursday (2 march) on messenger version 1.4, Among the changes announced (Messenger Profile API #690 and Messenger Code API #689) and many other cool stuff !

Glad you liked it :smile:

Just out of curiosity, why you needed one bot that supports many pages at once ?

in this case you need to call get_messenger_code as mush as the number of your pages.

jonchurch:
It's for a client, they are providing facebook bots to their customers
(people who run HUGE conferences) to provide interactions at scale.

In order to make it so code doesn't have to be written whenever a new bot
is added, I'm making botkit support any number of bots.
So essentially, one controller, any number of instances of bots, new bots
can be added or removed at any time.

It's similar to how botkit handles Slack Bots, one controller X number of
teams. Each team has some unique config stuff, but in the Facebook case,
each page's bot would have different content. Luckily Botkit Studio handles
the scripts for us, all we have to do is route messages to the right bots
etc.

So in this case, each page would have its own Messenger Code.

It's an edge case right now, but I think that it's totally logical to see a
future where Facebook Bots are "added" to a page similarly to how Slackbots
are currently "added" to a team. So one "app" is providing services to
multiple pages at a time.

jonchurch:
That being said, I'm giving your PR as it stands a big thumbs up :)

ouadie-lahdioui:
Interesting, thank you for sharing your use case here.

Issue #683, opened 2017-02-28
bot.say fails when using middleware (dashbot)


alecl:
Happening with 0.4.7 and latest 0.5.1. Worked fine before putting in a dashbot API key. My bot is based on the Slack starter from howdy.

I'm running this after controller.startTicking().
```

  var myTeam = 'Txxxxx';

  controller.storage.teams.get(myTeam, (err, team) => {
    if (err || !team) {
      controller.log.error('Error: could not load my team from storage system', err);
    } else {
      var botToken = team.bot.token;
      var channel = team.bot.createdBy;

      if (process.env.NODE_ENV === 'staging') {
        channel = 'Cxxxxx';
      }

      var startupBot = controller.spawn({
        token: botToken,
      });
      startupBot.say({ text: `started`, channel });
    }
  });
```

After hitting the startupBot.say link it then fails here where worker is null so .send isn't valid.

https://github.com/howdyai/botkit/blob/0188c5bfce2c7bf1942234d9aa057cc1c588ee61/lib/CoreBot.js#L1031

If I remove the DASHBOT_API_KEY all is fine again.

peterkinmond:
@alecl what repo are you testing this with? I tried to reproduce with my bot (using Botkit 0.4.7 and Dashbot 0.6.1) but didn't see the error. The `bot.say` function worked both with and without the middleware. Used the following:

```
var bot = controller.spawn({
  token: process.env.SLACK_TOKEN
});

if (process.env.DASHBOT_TOKEN) {
   // Add the dashbot middleware - used for bot analytics (dashbot.io)
  var dashbot = require('dashbot')(process.env.DASHBOT_TOKEN).slack;
  controller.middleware.receive.use(dashbot.receive);
  controller.middleware.send.use(dashbot.send);
}

bot.api.channels.list({}, function(err, response) {
  var channel = response.channels[0];
  bot.say({text: 'started', channel: channel.id});
});
```

alecl:
Super weird. I pulled https://github.com/howdyai/botkit-starter-slack and wasn't able to repeat. Then I messed with the versions of dashbot back and forth and now I can't get it to stop happening even on a fresh pull in a new directory. I found the root cause though.

First off there's a missing error check here which obscured the problem

https://github.com/howdyai/botkit/blob/master/lib/CoreBot.js#L1031

I added it like this and this is the error I am seeing when dashbot is enabled. Not sure if this is the error style that's appropriate. If so happy to PR it.
```
[...]

PR #691, opened 2017-03-05
Added error logging for say and spawn.run

alecl:
I got burned trying to debug #683 because the err was ignored and never logged. Added logging for it to save some poor future soul.


benbrown:
thank you, @alecl

PR #692, opened 2017-03-06
Update facebook messenger menu persistence (refering to last messenger update)

Raniazy:
"Most bots have a common “menu” or “settings” option, and we are now making this more consistent and easier to access with the launch of persistent menu. This eliminates the need for people to remember text commands and provides a great way to restart the flow. The menu can be accessed by an icon in the message composer, and will support up to five custom actions." - Facebook
Here's how it looks : 
persistencemenu


ouadie-lahdioui:
I slightly migrate from threadSetting to messenger profile API. take a look at #690

PR #694, opened 2017-03-06
Slack fix: properly ignore all messages with bot_id

### LINKED ISSUE #663: Old events firing when botkit restarts (after upgrade)
Hi. I just upgraded to latest version of botkit from a pretty old version (v0.1.1).  The problem is now when I start (or restart) botkit, it seems to be remembering the last message that was received, and is triggering that event again. The reason I know it's an old event, is the incoming message contains the same value, e.g ```{ reply_to: 106 }``` even between restarts.  The only way to make it forget 106 is to send a new message into slack. But then that new message becomes the one that gets stuck as `reply_to` when I restart botkit.

This was definitely not an issue in the previous version I was using.

My event handlers look mostly like this: (however the one that gets triggered will depend on the current message stuck in `reply_to`)

```javascript
controller.hears('\\b(subscribe)\\b', ['direct_message','direct_mention','mention'], (bot, message) => {
  // Respond code
});
```

jonchurch:
Filter out any message that comes in with `bot_id` on the body. Currently we are filtering out messages that have `message.user === bot.identity.id && message.bot_id` but as @sporkd pointed out, bot_id is not always present, so messages from our/other bots can slip through. 

This essentially makes our bots deaf to other bots, for now.

Extends #268 
Closes #663

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -725,7 +725,7 @@ function Slackbot(configuration) {
                     return false;
 
                 } else {
-                    if (message.user == bot.identity.id && message.bot_id) {
+                    if (message.user == bot.identity.id || message.bot_id) {
                         return false;
                     }
                     if (!message.text) {
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -707,7 +707,7 @@ function Slackbot(configuration) {
                     return false;
                 } else if (message.channel.match(/^D/)) {
                     // this is a direct message
-                    if (message.user == bot.identity.id && message.bot_id) {
+                    if (message.user == bot.identity.id || message.bot_id) {
                         return false;
                     }
                     if (!message.text) {


sporkd:
`&& message.bot_id` appears in another block. Do we also need to change it here? https://github.com/jonchurch/botkit/blob/d33fbfb16be51f78d40196124dbc5365561e90f3/lib/SlackBot.js#L728

jonchurch:
Thanks @sporkd 
It was just ignoring Direct Messages, that second block is to ignore Ambient, Direct Mention, and Mentions. Fixed now

Although this opened a new can of worms, should we be blocking ALL messages that have a bot_id? As in channel joins, file uploads, etc. Currently we are passing custom events like bot_channel_join but this is referrring to _our_ bot not _any_ bot as the wording suggests.

Essentially right now we don't have a protocol for handling, or reliably identifying bot messages from bots besides our own.

That's out of the scope of this PR, but that's a thought I keep running into when answering issues.

monte-hayward:
I was working around this using the following hack:

```
  if (msg.user === bot.identity.id || msg.user_profile && msg.user_profile.name === 'watson-mms') {
    log('mms-slack doesn\'t talk to itself');
    return;
  }
```

It prefers a positive id and falls back to name. Would love to see a merge.

monte-hayward:
@benbrown are we close to a merge on this one?

The workarounds are ok for simple listen & respond. 
Actual conversations (`startConversation` or `createConversation`) are blocked by this echo effect.
Thanks

benbrown:
I am working on a solution for this as part of this project:

https://github.com/howdyai/botkit/projects/6

benbrown:
@monte-hayward what are you trying to achieve? Ignore all messages from any bot?

@sporkd what about you?

jonchurch:
Don't think this is needed anymore

Issue #473, opened 2016-11-02
Receiving duplicate messages on bot.reply 


liornaar:
Hi,
Does anyone else encounter this issue?
Whenever i do bot.reply i get duplicate message:

![image](https://cloud.githubusercontent.com/assets/13891352/19935165/c19467b0-a121-11e6-9ef0-c248b2ac107d.png)

also in logs i see:

debug: CUSTOM FIND CONVO 1180150065384201 1180150065384201
debug: I HEARD [ 'hi', 'hello', 'bot', 'whats up', 'sup' ]
debug: No handler for stats:heard_trigger
debug: No handler for tick
debug: GOT A MESSAGE HOOK
debug: RECEIVED MESSAGE
debug: CUSTOM FIND CONVO 1180150065384201 1180150065384201
debug: I HEARD [ 'hi', 'hello', 'bot', 'whats up', 'sup' ]
debug: No handler for stats:heard_trigger
debug: WEBHOOK SUCCESS { recipient_id: '1180150065384201',
  message_id: 'mid.1478100520671:324459a102' }
debug: WEBHOOK SUCCESS { recipient_id: '1180150065384201',
  message_id: 'mid.1478100520939:d3d430cb89' }

when sending only one message...

agamrafaeli:
Can you submit a simple code snippet with a bot that performs this behaviour?

Also, which integration are you using? and what version?

liornaar:
Yes, Sorry for lack of details..
I'm using the bot with Facebook Messenger 
It happened since i started using Botkit.. even when I just pasted the code example from the readme file 
although im using my own express server

Botkit version: └── botkit@0.4.0

```
var Botkit = require('botkit');
var controller = Botkit.facebookbot({
        access_token: process.env.access_token,
        verify_token: process.env.verify_token,
})

var bot = controller.spawn({
});

//using my own server
server.listen(443, function () {
    console.log('gonna try to add web hook . . .');
    controller.createWebhookEndpoints(app, bot, function () {
        console.log('ONLINE!');
    });
});

// this is triggered when a user clicks the send-to-messenger plugin
controller.on('facebook_optin', function(bot, message) {

    bot.reply(message, 'Welcome to my app!');

});

// user said hello
controller.hears(['hello'], 'message_received', function(bot, message) {

    bot.reply(message, 'Hey there.');

});

controller.hears(['cookies'], 'message_received', function(bot, message) {

    bot.startConversation(message, function(err, convo) {

[...]

PR #695, opened 2017-03-07
Spawn only one bot per team

slambdin1:
**Related Issue**
https://github.com/howdyai/botkit/issues/473#issuecomment-284810694

**What does this PR do?**
This Pr makes sure that only one bot is spun up per team. The issue that I was seeing before was that when the bot was authorized twice to the team, then when the bot posted in a channel, it would post multiple times (Note: this does not happen in a conversation)

**Description of Changes**
`lib/SlackBot.js`: updated the `ouath access` method to check if a bot has already been spawned for that team. If one has, then it will _not_ spawn another bot. If a bot has not been spawned, then it _will_ spawn a new one and set up all of the necessary hooks.

**What gif most accurately describes how I feel towards this PR?**

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -533,26 +533,40 @@ function Slackbot(configuration) {
                                     };
                                 }
 
-                                var bot = slack_botkit.spawn(team);
 
-                                if (auth.incoming_webhook) {
-                                    auth.incoming_webhook.token = auth.access_token;
-                                    auth.incoming_webhook.createdBy = identity.user_id;
-                                    team.incoming_webhook = auth.incoming_webhook;
-                                    bot.configureIncomingWebhook(team.incoming_webhook);
-                                    slack_botkit.trigger('create_incoming_webhook', [bot, team.incoming_webhook]);
-                                }
+                                var bot;
 
-                                if (auth.bot) {
+                                spawned_bots.forEach(localBot => {
+                                    if (localBot.config.id && localBot.config.id == team.id) {
+                                        // WAIT! We already have a bot spawned
+                                        // here. Instead of using the new one,
+                                        // use the exist one.
+                                        bot = localBot;
+                                    }
+                                });
 
-                                    team.bot = {
-                                        token: auth.bot.bot_access_token,
-                                        user_id: auth.bot.bot_user_id,
-                                        createdBy: identity.user_id,
-                                        app_token: auth.access_token,
-                                    };
-                                    bot.configureRTM(team.bot);
-                                    slack_botkit.trigger('create_bot', [bot, team.bot]);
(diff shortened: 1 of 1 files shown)


jonchurch:
[This middleware function](https://github.com/howdyai/botkit/blob/master/lib/SlackBot.js#L30) _should_ prevent spawning multiple bots?

```javascript
    // Middleware to track spawned bots and connect existing RTM bots to incoming webhooks
    slack_botkit.middleware.spawn.use(function(worker, next) {

        // lets first check and make sure we don't already have a bot
        // for this team! If we already have an RTM connection, copy it
        // into the new bot so it can be used for replies.

        var existing_bot = null;
        if (worker.config.id) {
            for (var b = 0; b < spawned_bots.length; b++) {
                if (spawned_bots[b].config.id) {
                    if (spawned_bots[b].config.id == worker.config.id) {
                        // WAIT! We already have a bot spawned here.
                        // so instead of using the new one, use the exist one.
                        existing_bot = spawned_bots[b];
                    }
                }
            }
        }

        if (!existing_bot && worker.config.id) {
            spawned_bots.push(worker);
        } else if (existing_bot) {
            if (existing_bot.rtm) {
                worker.rtm = existing_bot.rtm;
            }
        }
        next();

    });

```

And then is [called in CoreBot.js](https://github.com/howdyai/botkit/blob/master/lib/CoreBot.js#L1034) when spawning bots like so:
```javascript
  botkit.spawn = function(config, cb) {


        var worker = new this.worker(this, config);
        // mutate the worker so that we can call middleware
        worker.say = function(message, cb) {
            botkit.middleware.send.run(worker, message, function(err, worker, message) {
                worker.send(message, cb);
            });
        };
        botkit.middleware.spawn.run(worker, function(err, worker) {

            botkit.trigger('spawned', [worker]);

            if (cb) { cb(worker); }

        });

        return worker;
    };
```
Perhaps it currently is not preventing spawning/connecting multiple bots? I assume that is the intended use of this middleware. 

If that was the case, and worker is not being mutated to be `existing_bot` outside of the middleware function's scope, we could fix it so spawn always returns the existing  bot from the middleware if present.

slambdin1:
@jonchurch based on the interaction that I've been seeing when I authorize the bot twice to the same time, the middleware is not preventing the spawning of multiple bots. 

What I _believe_ is happening is that the new bots are still spun up, and it uses the same RTM connection, but when an event gets triggered like `hears` it gets triggered on all of the bots so all of the bots end up responding to the event ( <-- This is a shot in the dark at the moment)

I did try messing around in the code snippets that you posted. For example: I tried  returning the existing bot when it existed, but duplicate messages still appeared. I almost think I need to destroy the worker if an existing bot exists?

Issue #681, opened 2017-02-27
response process is not ended - interactive message buttons wont work


codepenman:
![image](https://cloud.githubusercontent.com/assets/14257947/23385084/ace02c86-fd02-11e6-8c31-f3d957a9cddb.png)
 
res.end() is missing in the callback of /slack/receive route.

mirekm:
Any immediate solution to that or temp workaround? I can't get interactive message to work either: I double checked the interactive messages callback (from localtunnel.me) on Slack but it seem Slack can't connect. What's weird is that when I test it with something like `curl -X POST interactive_messages_callback_url` the endpoint seems to be accessible and available.

PR #697, opened 2017-03-08
Add responding with http success on slack slash command

windmark:
**Related issue**
#681 

**What does this PR do?**
It simply responds to the payload command, which indicates a slash command, with an HTTP 200. Without this, the slash command itself would still work but, Slack would timeout after 3000 ms due to the lack of a confirmation.

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -223,6 +223,7 @@ function Slackbot(configuration) {
                 return handleInteractiveMessage(payload, bot);
             } else if (payload.command) {
                 // this is a slash command
+                res.send('');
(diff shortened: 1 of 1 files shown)


benbrown:
Unfortunately, this breaks the way the replyPrivate feature of slash commands work.

If you do not want to use one of the built in reply features that are slash command related, you can do a bot.res.send('') in your own callback.

fletchrichman:
use replyAcknowledge when receiving a slash command: https://github.com/howdyai/botkit/blob/061126bc7bf78e2ffe1493d3d940430133a1ec12/docs/readme-slack.md#botreplyacknowledge

PR #699, opened 2017-03-08
Fixes slash commands when using internal webserver

mbensch:
The latest changes of Slack removed the concept of a bot from things like slash command. Botkit was still assuming that slack will send the `bot` key in the webhook payload which is not the case.

This is a simple fix that will remove the requirement for that information to be present for slash command, outgoing webhooks and interactive messages. The `handleEventsAPI` method will actually try to use that data so I still included it for that case but fail if we don't find it in the payload.

It was a blocker for us at message.io so we fixed and tested it in our own fork.
Should solve https://github.com/howdyai/botkit/issues/590#issuecomment-285186862


benbrown:
thank you @mbensch

Issue #687, opened 2017-03-02
no option to turn off default 'Initializing Botkit vXXX' message


mysticflute:
Version 0.5.1

CoreBot.js has a line that always logs to the console:

console.log('Initializing Botkit v' + botkit.version());

I think this should either use botkit.log, or there should be a config option to turn it off.

JoshuaKlassen:
Has anybody started to look into this? I would like to take a crack at it

JoshuaKlassen:
I have created a PR for this: #703

PR #703, opened 2017-03-12
Add option to disable default startup message

JoshuaKlassen:
This PR is to add the option to disable the `Initializing Botkit vXXX` message as per #687 

To achieve this, `CoreBot.js` was modified to check for `disable_startup_message` on the configuration object.
If `disable_startup_message=true`, the startup message will not be shown.

Example:
```
var controller = Botkit.slackbot({
    debug: true,
    disable_startup_message: true
});
```

<s>Upon getting started, I noticed that `.vscode` files are not in the `.gitignore` (I am using VSCode)
If this is intentional let me know and I can undo the change.</s>
My VSCode editor generated a `settings.json` file, I have added that to the `.gitignore` incase this is something that should be ignored - if not it can be removed.


jonchurch:
I'd suggest changing it from a `console.log` to a `botkit.debug` instead of adding a new config option to the growing list.

JoshuaKlassen:
Sounds good, I'll make the change to `botkit.debug`

As for the starter kits, would that be acceptable to make debug enabled by default?

peterswimm:
Could you explain why setting debug to being on default is preferred behaviour? My gut is that it would be an annoyance for most first time users, and is only really useful for people diagnosing unusual problems.

The initializing message is useful for us, particularly with regards to roadmap plans and supporting problems, and will eventually contain other useful information for developers. but being able to opt-out of receiving them is probably a reasonable request.

JoshuaKlassen:
I just did what @jonchurch suggested, I have no preference

benbrown:
I like @JoshuaKlassen's original idea - allow the message to be disabled independently of debugging.  I believe this IS important to be on by default.

JoshuaKlassen:
I reverted back to original idea. I also felt it would be appropriate to make the config option plural, seeing as how we could add more startup messages in the future

jonchurch:
👍 Sounds good to me. 

@peterswimm I agree, that would likely annoy developers. I was suggesting a scenario where the startup is on by default, but dependent on debug mode. 

Y'alls solution is cleaner and more predictable though.

JoshuaKlassen:
@peterswimm, @jonchurch, @benbrown: is there anything I am missing, or is this good to go?

Issue #674, opened 2017-02-24
Simulate typing in Conversations


AbhiSivasailam:
I know its possible to replyWithTyping, etc...But how can typing be simulated in conversations, say, in the context of a _convo.say()_

devhaoy:
This should be labelled "enhancement required" rather than "help wanted". convo.say / ask should definitively feature the typing option

thierryskoda:
+1

altais:
Also interested 👍

AdanTL:
+1

nischayv:
I can work on this feature!

lorenzhk:
+1

jonchurch:
First lemme say, I don't like how most bot devs are using the typing indicator in their apps! 

People keep writing articles about how it "improves UX" but I don't think enough people have tested it for real for it to be a "rule", or to know when users just want us to get to the point. It annoys me because it ends up increasing the ask of your user's time for basically a spinning beach ball.

Also when the bot types for _the same amount of time_ (perceptually) for messages of varying length, it really breaks the whole illusion for me and is actually distracting. (disclosure: I'm a millenial, I read intently into typing durations whenever I encounter them with humans)

And botkit's typing delay is too slow in my opinion, so if y'all are gonna jump on the typing train you should think through how long the bot should type and adjust the duration until it's fixed. 

**THAT BEING SAID! Here is a one line change that will give you what you want in the meantime**

In lib/CoreBot.js change [line 632](https://github.com/howdyai/botkit/blob/master/lib/CoreBot.js#L632) from `this.task.bot.reply`to `this.task.bot.replyWithTyping`. And that should do it! All messages sent via conversation will now use typing indicators.

It's a hack, it's not tested, but it's working for me. replyWithTyping is just a wrapper around reply that first sends a typing indicator, sets a timeout, and then sends the message after the timeout. Mess with replyWithTyping to change the typing time. 

This requires maintaining your own fork of botkit, btw, until someone makes a PR for this. I will get on it tmr probably.

KarlisJ:
@jonchurch I know this is old but I wanted to give my opinion. I do agree with you that typing indicators are annoying most of the times. However I dont think should be all on the developers.

I dont know how it is with slack but I hate that for messenger bot you have to send typing indicator as single message. If developers could attach typing: "on" to message we could actually build better flows. 

[...]

PR #710, opened 2017-03-15
Add functionality for typing in conversations

nischayv:
Issue - #674

This PR allows users to simulate typing in conversations, similar to what is being done with replyWithTyping. 
```
bot.startConversation(message, (err, convo) => {
        convo.askWithTyping('Are you sure?', [  {
                pattern: bot.utterances.no,
                default: true,
                callback: (response, convo) => {
                    convo.sayWithTyping('*Phew!*');
                    convo.next();
                }
            }
       ]);
});
```

I followed what was being done for replyWithTyping.

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -191,6 +191,10 @@ function Botkit(configuration) {
             this.addMessage(message);
         };
 
+        this.sayWithTyping = function(message) {
+            this.addMessage(message, null, true);
+        };
+
         this.sayFirst = function(message) {
             if (typeof(message) == 'string') {
                 message = {
@@ -245,7 +249,7 @@ function Botkit(configuration) {
             return;
         };
 
-        this.addQuestion = function(message, cb, capture_options, thread) {
+        this.addQuestion = function(message, cb, capture_options, thread, typing) {
             if (typeof(message) == 'string') {
                 message = {
                     text: message,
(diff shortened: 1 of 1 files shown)


devhaoy:
Hi @nischayv  I found several issues with your fix. 
1/ All convo.ask which `message` is not a string (buttons, quick_replies etc.) won't fire and give error `API ERROR { message: '(#100) Message cannot be empty, must provide valid attachment or text'`
2/ When sending attachment with convo.say, throw error `TypeError: Cannot read property 'length' of undefined`
3/ Sending emojis (with no text) messes up the whole convo. Probably a a shortcoming of the `length` property. Surprisingly, the emoji will be ignored and fired in the subsequent thread. This seems to confuse botkit, and eventually all the convo.say('text') are sent in the wrong order. 

@jonchurch  are you working on your own fix to simulate typing in convo?

nischayv:
@devhaoy Thanks for pointing that out. I'll look into what's going wrong. By any chance do you know why that's happening? Any help would be appreciated. I'm new to botkit, and would really like to contribute to it.

devhaoy:
@nischayv  I have little understanding of the Corebot mechanism, I'd rather not mislead you! As Jon Church pointed out it is some kind of hack to tweak `bot.replyWithTyping`  for use with convo, so I suppose `typing` needs to be an isolated argument dedicated to convo.

jonchurch:
@devhaoy This is a different fix than the hack I proposed in #674 

I'm just switching out the bot.reply that is used to finally resolve sending a message in a convo (in CoreBot.js) with the existing bot.replyWithTyping, a function that is functionally the same as reply but wraps it with a user_typing call first. Mine will _definitely_ break if you are using a platform that doesnt have replyWithTyping implemented (like twilio, I believe).

I haven't started on a PR yet, but I was going to check a config variable on the controller to decide between using bot.reply or bot.replyWithTyping in CoreBot.js

k001:
What is the differences between #724 and this?
Both wants the same behavior.

nischayv:
@k001 The main difference seems the manner of usage. In my PR, we can do convo.replyWithTyping, which is similar to bot.replyWithTyping. In the other PR, it would have to be an added property with_typing. Given that both PR's did not account for platforms where typing is not available(twilio maybe), it may break things. I'll update my PR this weekend and make sure this feature is working.

jonchurch:
@k001 The difference is that #724 will force all conversations to have typing on by default, within slack. 

This creates a new function which you can explicitly use instead of `convo.say()`

c0nstructer:
@jonchurch what's the status with this merge and feature?

PR #716, opened 2017-03-19
Fix broken facebook documentation links.

jkukul:


peterswimm:
Thanks for fixing that!

Issue #715, opened 2017-03-18
groups.replies, im.replies missing


joseym:
It seems as tho only `channels.replies` are currently supported.

https://github.com/howdyai/botkit/blob/3b175b68cc5cdf6ae7e9d075cfefcd697a8ed965/lib/Slack_web_api.js#L38

However, the slack web api also has methods for `groups.replies` and `im.replies`.

benbrown:
I merged your PR. It'll make it into the next version published to NPM.

PR #720, opened 2017-03-20
Add support for *.replies

joseym:
As indicated by #715, the `Slack_web_api.js` file is missing the replies method on all except `channels`.


benbrown:
Thanks @joseym

PR #723, opened 2017-03-21
Change behavior of conversation timeouts

benbrown:
This change seeks to disambiguate the difference between similarly named components of the conversation system. 

Prior to this change, when a conversation timed out, it would look for and automatically run a thread called "timeout."  You could also specify "timeout" as an action -- this would cause the conversation to end and be marked as timeout, but did _not_ trigger the timeout branch.

To resolve this, the timeout branch has been renamed to "on_timeout"

Developers who have previously used this undocumented feature will have to rename their timeout branch.

Documentation for this feature has been added to the readme, and can be viewed here:
https://github.com/howdyai/botkit/pull/723/files#diff-0730bb7c2e8f9ea2438b52e419dd86c9R919


benbrown:
This PR also adds the option of specifying a handler for the timeout event using `convo.onTimeout(handler)`

lib/CoreBot.js
```diff
@@ -553,6 +557,10 @@ function Botkit(configuration) {
             return outbound;
         };
 
+        this.onTimeout = function (handler) {
+            this.timeOutHandler = handler;
+        }
```
anonrig: Not important but since linting is important: missing semicolon (`;`)

lib/CoreBot.js
```diff
@@ -568,10 +576,13 @@ function Botkit(configuration) {
                         (duration > this.task.timeLimit) && // timelimit is up
                         (lastActive > this.task.timeLimit) // nobody has typed for 60 seconds at least
                     ) {
-
-                        if (this.threads.timeout) {
+                        // if timeoutHandler is set then call it, otherwise follow the normal flow
+                        // this will not break others code, after the update
+                        if (this.timeOutHandler != null || this.timeOutHandler != undefined) {
```
anonrig: A shorter version could be `if (this.timeOutHandler)` since we're only checking for non null and defined conditions.

lib/Studio.js
```diff
@@ -512,7 +512,6 @@ module.exports = function(controller) {
                     json = JSON.parse(body);
                 } catch (e) {
                     console.log('Error in Botkit Studio Stats:', e);
-                    return reject('Invalid JSON');
```
anonrig: Can you elaborate on the reason of this removal @benbrown ?

benbrown: @anonrig the reject is still happening right outside the catch. we've seen weird behavior when calling callbacks inside try/catch blocks.

lib/CoreBot.js
benbrown: Fixed!

lib/Studio.js
anonrig: Ah, I see.  Thanks for the feedback!

lib/CoreBot.js
anonrig: Also shouldn't this line be: `this.timeOutHandler != null && this.timeoutHandler != undefined` if it is null OR undefined, next line will fail. @benbrown

lib/Studio.js
benbrown: @anonrig thank you!

lib/CoreBot.js
benbrown: @anonrig what about using typeof?

anonrig: Why would you want to check the type of a variable? I didn't understand. In JS `null` or `undefined` is transformed into `false` in logical operator check. Therefore `if (this.timeOutHandler)` will cover both of the cases in this if statement.

benbrown: Just to make sure it is a valid function before calling it...

What I did: added a test on the typeof(handler) when _setting_ the handler.

PR #747, opened 2017-04-03
provide definition of RTM

sheldonrampton:
This readme makes numerous references to "RTM" but never explains that this stands for Slack's Real Time Messaging API. This pull request adds that reference, including a link to Slack's Real Time Messaging API documentation.


peterswimm:
Thanks! This doc is a remnant from when RTM was the major way to create a bot, so needs some rewriting. This is a good first step!

PR #748, opened 2017-04-04
encrypt team and user tokens before storing to db

abratnap:
This is to provide functionality to encrypt the token before we store it in underline botkit storage.

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -4,6 +4,7 @@ var express = require('express');
 var bodyParser = require('body-parser');
 var querystring = require('querystring');
 var async = require('async');
+var Cryptr = require('cryptr');
 
 function Slackbot(configuration) {
 
@@ -77,6 +78,10 @@ function Slackbot(configuration) {
                 slack_botkit.config.scopes = slack_app_config.scopes.split(/\,/);
             } else {
                 slack_botkit.config.scopes = slack_app_config.scopes;
+            }
+            if (slack_app_config.encrypt && slack_app_config.encryptKey) {
+                slack_botkit.config.encrypt = slack_app_config.encrypt;
+                slack_botkit.config.cryptr = new Cryptr(slack_app_config.encryptKey);
             }
             if (cb) cb(null);
         }
@@ -335,7 +340,13 @@ function Slackbot(configuration) {
      * ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ */
 
     slack_botkit.saveTeam = function(team, cb) {
-        slack_botkit.storage.teams.save(team, cb);
+        var aTeam = Object.assign({}, team);
+        if (slack_botkit.config.encrypt) {
+            aTeam.bot.token = slack_botkit.config.cryptr.encrypt(team.bot.token);
+            aTeam.bot.app_token = slack_botkit.config.cryptr.encrypt(team.bot.app_token);
+
+        }
+        slack_botkit.storage.teams.save(aTeam, cb);
     };
 
     // look up a team's memory and configuration and return it, or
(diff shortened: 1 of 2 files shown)


lib/SlackBot.js
```diff
@@ -4,6 +4,7 @@ var express = require('express');
 var bodyParser = require('body-parser');
 var querystring = require('querystring');
 var async = require('async');
+var Cryptr = require('cryptr');
```
anonrig: I would personally suggest using bcrypt if you don't want to use any native crypto library.

abratnap: thank you @anonrig. I'll look into that.

tmundt: @abratnap For what purpose do you need to encrypt the team and user tokens? I get the basic idea (security measures), but what if you would like to read those data manually, lets say for checking reasons what team has connected to your app? Do you provide a decrypt function to establish connection of your bot to each connected teams?

abratnap: @tmundt the purpose of this is to securely store tokens in the backend store. Any compromise to the backed store should not leak tokens. The user can give a key in a configuration which can be used to encrypt tokens before storing. The same key can be used to decrypt the token. The decrypt function in same library like `Cryptr` can be used for that.

anonrig: The problem with this approach is `cryptr` uses AES as default. Due to AES, and it's algorithm if the database is compromised, yes the data can't be easily decrypted. But assume that a hacker has compromised your database instance, if that is the case, it's most likely that he/she may have access to your node instance also.

anonrig: If you want something secure enough, than this approach is not the perfect method. It's just a precaution that in the end, it's not a blocker for the hacker.

abratnap: @anonrig In my case, I'm using hosted DB solution. Compromised DB will have a severe impact as I have data for few hundred teams (And thousands of users). I believe I can protect the encryption key (on app side) better than the database storage/access. 

I know it's a precaution so wanted to keep it optional.

anonrig: In that case can you use a more widely known encryption library like `bcrypt` please?

lib/SlackBot.js
```diff
@@ -335,7 +340,13 @@ function Slackbot(configuration) {
      * ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ */
 
     slack_botkit.saveTeam = function(team, cb) {
-        slack_botkit.storage.teams.save(team, cb);
+        var aTeam = Object.assign({}, team);
```
anonrig: Using Object.assign will break the applications running Node < 4.

lib/SlackBot.js
```diff
@@ -511,7 +522,7 @@ function Slackbot(configuration) {
                     var scopes = auth.scope.split(/\,/);
 
                     // temporarily use the token we got from the oauth
-                    // we need to call auth.test to make sure the token is valid
+                    // we need to call auth.test to make sure the bot.app_ is valid
```
anonrig: Why did you change this?

lib/SlackBot.js
```diff
@@ -614,7 +625,12 @@ function Slackbot(configuration) {
 
                                             // Always update these because the token could become invalid
                                             // and scopes could change.
-                                            user.access_token = auth.access_token;
+                                            if (slack_botkit.config.encrypt) {
```
anonrig: More appropriate and readable format is: `slack_botkit.config.encryptionEnabled`

lib/SlackBot.js
```diff
@@ -78,6 +79,10 @@ function Slackbot(configuration) {
             } else {
                 slack_botkit.config.scopes = slack_app_config.scopes;
             }
+            if (slack_app_config.encrypt && slack_app_config.encryptKey) {
+                slack_botkit.config.encrypt = slack_app_config.encrypt;
```
anonrig: Why do we need this line? Can you explain.

lib/SlackBot.js
tmundt: @abratnap I encrypt other data in the storage database being generated by botkit in a separate table, I use the slack_id as key.

PR #754, opened 2017-04-05
Add `beforeThread` hooks

benbrown:
This adds a new hook for firing handler functions in between conversation threads.

This is useful for doing asynchronous reads/writes in the middle of a conversation, such as writing user responses to a database before displaying a confirmation message.


devhaoy:
HI @benbrown  would you mind adding an example to this fix?

benbrown:
@devhaoy what specifically would you like to see an example of? There are some examples of the beforeThread hooks in the readme files.

devhaoy:
@benbrown  sorry I hadn't seen the readme update, it's clear now. The fix will only work for bots using the thread navigation system if I understand correctly? That's missing most of botkit powered bots... any mean to make the fix work within the default thread?

devhaoy:
hi @benbrown  any advice how to get the same result while staying in the default thread? Not using threads because I feel the code looses in clarity, I'm simply calling a function and convo.next to progress in the conversation.

PR #761, opened 2017-04-11
Dev/newdocs

puneetmathurDS:
Specifically, the use of "port" for the environment variables in the examples that contain lines such as this one:
if (!process.env.clientId || !process.env.clientSecret || !process.env.port) {
Please change "port" to the more commonly used "PORT" variable.

the above change will be implemented.


peterswimm:
You should make changes off the master branch, the newdoc branch was just an experiment on my end.

Issue #768, opened 2017-04-13
Slackbot_worker throws Error when Slack Select menu is used


clintonhalpin:
Testing out the new Slack Select menu's using a conversation. This line throws an error when sent a slack select response [Slackbot_worker.js#L741](https://github.com/howdyai/botkit/blob/master/lib/Slackbot_worker.js#L741)

```javascript
message.text = message.text.replace(direct_mention, '').replace(/^\s+/, '').replace(/^\:\s+/, '').replace(/^\s+/, '');
```
Error:
```
TypeError: Cannot read property 'replace' of undefined
    at Object.bot.findConversation
```
This is because `message.text` is not there and instead the response is in `message.actions[ 0 ].select_values`

[Gist of Message Response](https://gist.github.com/clintonhalpin/6245ff52257c16f5e7bebf2be714521f)

Details:
```
Version: 5.2
OS: OSX 10.3, Ubuntu 16.4
```

matt-oakes:
This should be closed now I think.

clintonhalpin:
@matt-oakes Yep! I tested this yesterday it it worked for me

PR #769, opened 2017-04-13
Make Interactive Slack Menus Work - fixes for #768

clintonhalpin:
Added a fix for #768. As far as I know slack menu's don't support multiple select values so this seems to work. There might be a smarter way to handle this case I'm just not totally privy to all of the workings in the library.

### DIFF
--- a/lib/Slackbot_worker.js
+++ b/lib/Slackbot_worker.js
@@ -738,6 +738,16 @@ module.exports = function(botkit, config) {
 
                         var direct_mention = new RegExp('^\<\@' + bot.identity.id + '\>', 'i');
 
+                        if( message.actions ) {
+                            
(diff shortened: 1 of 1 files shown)


Angry-Potato:
@clintonhalpin any workarounds for this before it gets merged? Tried setting the text prop on absolutely everything, no dice..

clintonhalpin:
@Angry-Potato I am using a fork of this repo in production at the moment `git://github.com/clintonhalpin/botkit.git#bug#768/getValueFromSelectConvo`

I don't think there is a workaround.

Angry-Potato:
cheers!

benbrown:
A fix for this is present in 0.5.5!

Issue #772, opened 2017-04-13
FacebookBot startTicking is not called unless setupWebserver is called


smalltalk-josh:
I am using botkit with Facebook Messenger, but I am not calling `setupWebserver`. My setup is similar to [using an Express web server alongside BotKit for Facebook Messenger](https://github.com/howdyai/botkit/blob/master/docs/readme-facebook.md#use-botkit-for-facebook-messenger-with-an-express-web-server), but I do call `createWebhookEndpoints`.

The biggest issue I noticed was that conversation did not work at all. See: [#503 - convo.say not working](https://github.com/howdyai/botkit/issues/503#issuecomment-270219099)

After digging around, I see that `startTicking` was not called, because I never called `setupWebserver`. This is a bug, since `startTicking` is a core to the bot, but `setupWebserver` is optional.

I am going to submit a PR, but the different bots handle calling `startTicking` differently.

`startTicking` - when called:

- ConsoleBot - [inside `text_botkit.listenStdIn`](https://github.com/howdyai/botkit/blob/master/lib/ConsoleBot.js#L72)
- BotFramework - [inside `bf_botkit.createWebhookEndpoints`](https://github.com/howdyai/botkit/blob/b13e5de9e208c7a76d59bf194ffd76f82a4e398c/lib/BotFramework.js#L156)
- Twiliobot - [inside `twilio_botkit.createWebhookEndpoints`](https://github.com/howdyai/botkit/blob/e00448dd2ee7d1bd048ebbfb3f3f3dd8f4e59026/lib/TwilioIPMBot.js#L221)
- Tropo - [inside ctor `function Tropobot(configuration)`](https://github.com/howdyai/botkit/blob/5984014fb5a988d2dd6e0bd47259d9f2c0a5e11b/lib/Tropo.js#L214)
- Sparkbot - [inside ctor `function Sparkbot(configuration))`](https://github.com/howdyai/botkit/blob/0188c5bfce2c7bf1942234d9aa057cc1c588ee61/lib/CiscoSparkbot.js#L477)
- Slackbot_worker - [inside - `bot.rtm.on('open')`](https://github.com/howdyai/botkit/blob/34b44646e95d0abff7dbc96d8a9acbce36f11878/lib/Slackbot_worker.js#L215)
[...]

Issue #503, opened 2016-11-21
bot.startConversation() returns No handler for conversationStarted


micmacys:
I have written a simple Facebook bot application using NodeJS and Express but every time I call bot.StartConversation() I get the following output in my server logs:
`debug: No handler for conversationStarted`

I looked over the code in CoreBot.js and it is triggering an event for conversationStarted.
`this.activate = function() {
            this.task.trigger(conversationStarted, [this]);
            this.task.botkit.trigger(conversationStarted, [this.task.bot, this]);
            this.status = active;
        };`

Is there a way to register a listener for conversationStarted which is not documented?

benbrown:
You can register an event for this, and it will stop logging.

Or, you can set debug: false in your controller options, and you will stop seeing this type of error.

miketen:
I tried registering the event but that doe not seem to work, seems like a "task" is trying to fire a conversationStarted event which does not exist. Botkit.trigger seems to be working, its not giving me an error in the console. I traced the error back to CoreBot.js.
`this.activate = function() {
          this.task.trigger('conversationStarted', [this]);
          this.task.botkit.trigger('conversationStarted', [this.task.bot, this]);
          this.status = 'active';
        };`

The example app i am trying to run can be found here, https://github.com/mvaragnat/botkit-messenger-express-demo. Notice how the example app does not provide code for starting a conversation which leads me to believe the conversation functionality is not supported for Facebook Messenger. I went ahead and added a conversation method to the example app and thats when I start to see No handler for conversationStarted errors.

benbrown:
What you are seeing is a debug message informing you that, though a conversationStarted event fired, you did not provide a handler for it.

This happens for _all_ of the events that fire, just to show you what is happening.

You can turn off these messages by passing debug: false into your controller configuration. Or, you can safely ignore them.

miketen:
After turning off debug mode I still can not get a conversation started. Can someone run the code below to see if it fails on startConversation?

**Code** :
`'use strict'

var express = require('express')
var bodyParser = require('body-parser')
var request = require('request')
var app = express()

var Botkit = require('botkit')

var token = "";

var controller = Botkit.facebookbot({
  //debug: true,
  access_token: token
})

var bot = controller.spawn({})

app.set('port', (process.env.PORT || 5000))

// Process application/x-www-form-urlencoded
app.use(bodyParser.urlencoded({extended: false}))

// Process application/json
app.use(bodyParser.json())


// for Facebook verification
[...]

PR #773, opened 2017-04-13
#772 move subscribed_apps and startTicking into createWebhookEndpoints

smalltalk-josh:
Resolve issue in #772 and #503 related to Facebook bot.
* `startTicking` is never called when setting up express outside of  `setupWebserver`


jonchurch:
Thank you for contributing! 

This fix is unnecessary in my opinion, as it does not solve a problem or bug that needs code changes.

see #772 for info about calling startTicking when using your own express server

If you'd like to submit a PR updating the FB docs in a way that you think clears up this confusion, that would be more likely to see a merge.

Issue #774, opened 2017-04-15
webpack: botkit.version() throws Error


monte-hayward:
webpack, having bundled code, makes botkit's package.json unreachable.
this causes the attempt to read version number to throw a runtime Error.

```
                var pkg = fs.readFileSync(__dirname + '/../package.json');
```
https://github.com/howdyai/botkit/blob/7e0afb49e8faa3e9abafa8f68e2a0c6a77a216b7/lib/CoreBot.js#L1188

botkit version: 0.5.2
webpack: 2.4.1
OS X: 10.12.4

```
fs.js:641
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
                 ^

Error: ENOENT: no such file or directory, open '/package.json'
    at Error (native)
    at Object.fs.openSync (fs.js:641:18)
    at Object.fs.readFileSync (fs.js:509:33)
    at Object.Botkit.botkit.version (/path/to/dist/bundle.js:23131:26)
    at Botkit (/path/to/dist/bundle.js:23184:50)
    at Object.Slackbot (/path/to/dist/bundle.js:103440:24)
    at Object.configure (/path/to/dist/bundle.js:84384:31)
    at Object.<anonymous> (/path/to/dist/bundle.js:99606:28)
    at __webpack_require__ (/path/to/dist/bundle.js:20:30)
    at /path/to/dist/bundle.js:66:18
```

jonchurch:
@peterswimm 🚫

PR #775, opened 2017-04-15
#774 expose package version to bundle

monte-hayward:
reads current package version in a bundler-detectable way.


monte-hayward:
thanks!

PR #776, opened 2017-04-15
Update readme-facebook.md

pedroalpha:
It's not node facebook_bot.js


benbrown:
What are you referring to here? The included sample is definitely called `facebook_bot.js` not `facebook-bot.js`

Issue #785, opened 2017-04-18
When a RegExp is not in an Array in `hears` does not work


alsuga:
The first does not work but the second does
`
const yesRegex = /^(yes|yep)/i;
controller.hears(yesRegex, [direct_message], (bot, message) => {
  bot.reply(message, 'yes');
});
const noRegex = /^(no|nop)/i
controller.hears([noRegex], [direct_message], (bot, message) => {
  bot.reply(message, 'no');
});
`

jonchurch:
@peterswimm 🚫

PR #786, opened 2017-04-18
#785 hears work when a Regex is not in an Array in `hears`

alsuga:


alsuga:
I want to resolve #785

benbrown:
good fix @alsuga thanks!

PR #790, opened 2017-04-20
Add FaceBook parametric codes

ouadie-lahdioui:
Yo,

Facebook just added (two days ago) the possibility to generate codes with ref parameters, So, the aim of this PR is to change on how we handle ```get_messenger_code``` to add support of parametric Codes.

Enjoy,


benbrown:
Great!!

PR #800, opened 2017-04-25
added facebook bot multi tenancy mode

alamusi:
- support access_token as object type (i.e. {"page_a_id":"page_a_token", "page_b_id":"page_b_token"} )
- backward compatible with current plain string page_token input, as of v0.5.2
- facebook_bot.js example updated

### DIFF
--- a/facebook_bot.js
+++ b/facebook_bot.js
@@ -27,6 +27,7 @@ This bot demonstrates many of the core features of Botkit:
 
     app_secret=<MY APP SECRET> page_token=<MY PAGE TOKEN> verify_token=<MY_VERIFY_TOKEN> node facebook_bot.js [--lt [--ltsubdomain LOCALTUNNEL_SUBDOMAIN]]
 
+  Use page_token='{<"PAGE_A_ID":"PAGE_A_TOKEN","PAGE_B_ID":"PAGE_B_TOKEN"}' to enable multi tenant
   Use the --lt option to make your bot available on the web through localtunnel.me.
 
 # USE THE BOT:
@@ -99,10 +100,18 @@ if(ops.lt === false && ops.ltsubdomain !== null) {
     process.exit();
 }
 
+var access_token = process.env.page_token;
+try {
+    var access_token = JSON.parse(access_token);
+    console.log("running on multi tenant mode", access_token);
+} catch (error) {
+    console.log("running on single tenant mode", access_token);
+}
+
 var controller = Botkit.facebookbot({
     debug: true,
     log: true,
-    access_token: process.env.page_token,
+    access_token: access_token,
     verify_token: process.env.verify_token,
     app_secret: process.env.app_secret,
     validate_requests: true, // Refuse any requests that don't come from FB on your receive webhook, must provide FB_APP_SECRET in environment variables
--- a/lib/Facebook.js
+++ b/lib/Facebook.js
@@ -127,7 +127,7 @@ function Facebookbot(configuration) {
             }
 
             //Add Access Token to outgoing request
-            facebook_message.access_token = configuration.access_token;
+            facebook_message.access_token = (typeof (configuration.access_token) === 'string') ? configuration.access_token : configuration.access_token[message.page];
 
             request({
                 method: 'POST',
@@ -159,6 +159,7 @@ function Facebookbot(configuration) {
         bot.startTyping = function(src, cb) {
             var msg = {};
             msg.channel = src.channel;
+            msg.page = src.page;
             msg.sender_action = 'typing_on';
             bot.say(msg, cb);
         };
@@ -166,6 +167,7 @@ function Facebookbot(configuration) {
         bot.stopTyping = function(src, cb) {
             var msg = {};
             msg.channel = src.channel;
+            msg.page = src.page;
             msg.sender_action = 'typing_off';
             bot.say(msg, cb);
         };
@@ -206,7 +208,7 @@ function Facebookbot(configuration) {
             }
 
             msg.channel = src.channel;
-
+            msg.page = src.page;
             bot.say(msg, cb);
         };
 
@@ -407,17 +409,18 @@ function Facebookbot(configuration) {
                 if (cb) { cb(null, facebook_botkit.webserver); }
             });
 
-
-        request.post('https://' + api_host + '/me/subscribed_apps?access_token=' + configuration.access_token,
-            function(err, res, body) {
-                if (err) {
-                    facebook_botkit.log('Could not subscribe to page messages');
-                } else {
-                    facebook_botkit.debug('Successfully subscribed to Facebook events:', body);
-                    facebook_botkit.startTicking();
-                }
-            });
-
+        var access_tokens = (typeof (configuration.access_token) === 'string') ? [configuration.access_token] : configuration.access_token;
+        Object.keys(access_tokens).forEach(function(page) {
+            request.post('https://' + api_host + '/me/subscribed_apps?access_token=' + access_tokens[page],
+                function(err, res, body) {
+                    if (err) {
+                        facebook_botkit.log('Could not subscribe to page messages with access_token', access_tokens[page]);
+                    } else {
+                        facebook_botkit.debug('Successfully subscribed to Facebook events with access_token', access_tokens[page], body);
+                    }
+                });
+        });
+        facebook_botkit.startTicking();
         return facebook_botkit;
 
     };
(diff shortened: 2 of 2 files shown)


ouadie-lahdioui:
Hello,

Can you share with us your use case ? why you need the multi tenancy mode ?

alamusi:
Hi,

We built a bot that runs in a SaaS model. To minimize the operational
overhead, we have one end point for all our clients' Facebook pages to
subscribe to.

The server behind the senses runs the same logic. Depending on which page
the message comes from, the bot responds according to client's specific
configuration.

Thanks,
Alamusi

jonchurch:
I think this is too specific, a good idea to maintain as a fork for those who want to solve this problem in this specific way. 

Im not sure how this would directly benefit users at this time. I have modified a version of botkit to support multiple teams, but in a dynamic way where your bot doesnt have to know anything about config options. It goes looking for them after receiving a request. Allows you to scale up more elastically imo. 

That I think is a more general purpose approach, if no token on the controller's config, hit a source of truth with a request and grab the config to spawn the bot to pass into `receiveMessage()`.

goodpic:
Hello, I am also running multiple bots on a single instance (server). I am currently doing so by loading and spawning bots from the multiple env files (they could be also in DB) as following.  
I am wondering whether there is a better or efficient way to achieve it, and though multi tenancy mode could help such a use case?
Or do you have any other recommended way to run multiple bots on the single instance/source code?

Load env files for each bot in a botRouter.js

```javascript
// Load multiple bot instances defined by each .env.BOT_NAME
const bots = ['bot-1', 'bot-2'];
const facebookHandler = [];
bots.forEach((bot) => {
  facebookHandler[bot] = require('../controllers/botkit')(bot);
});
```

Spawn them in a botkit.js

```javascript
module.exports = (name) => {
  console.log(`ID : ${name}`);
  // Create the Botkit controller, which controls all instances of the bot.
  env(path.join(appDir, `./.env.${name}`), { overwrite: true });
  if (!process.env.page_token) {
    console.log('Error: Specify a Facebook page_token in environment.');
    process.exit(1);
  }
  if (!process.env.verify_token) {
    console.log('Error: Specify a Facebook verify_token in environment.');
    process.exit(1);
  }

  const controller = Botkit.facebookbot({
    debug: true,
    receive_via_postback: false,
    require_delivery: true,
    verify_token: process.env.verify_token,
    access_token: process.env.page_token,
    studio_token: process.env.studio_token,
    studio_command_uri: process.env.studio_command_uri,
  });

  const bot = controller.spawn({});
```

alamusi:
@jonchurch could you share your solution (multiple teams) in details with us? it would be great that we don't have to maintain a fork that requires manually conflict resolutions/updates every time there is a new botkit version out.

@goodpic thanks a bunch for the sample code. one quick question - do your bots have different business logics (more as individual bots reside on the same server) or identical business logics (more as same bot serving different clients in a data driven way)? i like the multi tenancy being handled at bot level. it feels cleaner. just not sure how much overhead (memory/cpu) is to spawn every client instance (i.e. new facebook page subscription), if you ever tested it?

jonchurch:
@alamusi My fork is still under development, and not fully implemented (client work, so I'm working at their pace) but here is a comparison as it stands now, for development I am passing in my config objects manually, and looking them up right before spawning a bot for a page. This will be done with an api/db call (we are figuring out the specifics for their stack, but essentially is the same either way) when handling a page's messages. 

It is also a facebook specific fork, so I removed other connectors and most of their dependencies. [This fork is an incomplete example of multiple facebook bots on one server](https://github.com/howdyai/botkit/compare/master...jonchurch:multipleFB). The relevant changes are in Facebook.js, where I am spawning a bot per FB page. FB batches requests to your webhook, so it could have multiple pages coming at you in one request, so I spawn a bot once for the page when handling requests coming in. 

My client will be using Botkit studio to manage an elastic number of bots, so the "business logic" on my server is fairly generic, you can share validation middlewares, certain templated conversations/interactions (like "sign_up_for_notification", the logic for which is shared by all bots on my server, intentionally). When using Botkit Studio the way my client is, your bot server basically just needs to know what page and studio account it's dealing with per request, and send a request to Botkit Studio to get a conversation, if one has been triggered. 

Open to any questions (or criticisms), be nice it's in development still 😝

goodpic:
Hi @alamusi ,
My use case is "same bot serving different clients in a data driven way", and each client can sign-in to the system to add their own contents. I just got into the facebook messenger bot development recently but my gut feeling is that "one bot (webhook URL) per one facebook page (client)" could make things simpler. 
I haven't tested the memory/cpu usage of each bot, and that is exactly my concern. Though if I have each different webhook URL per bot (client), it should be fairly easy to scale out by load balancing based on the URL to the multiple server instances.

alamusi:
@jonchurch your fork appears to have quite a lot of changes across the board! passing in bot config is certainly one neat way of solving the problem... just being curious, how do you manage the interim when you have to rebase/merge the botkit releases? 

the reason I chose to make the simplest possible update was to minimize the overhead to upgrade with botkit until my pull request is approved (or someone else comes up with a better solution so I can migrate over) 😄

alamusi:
@goodpic If i understand your solution right, you have one facebook page, one facebook app, and one server end point per client. the multi tenancy you built was to allow multiple server end points to reside on the same server, correct?

my case is slightly different, what i like to have is multiple facebook pages to consume exactly the same facebook app, and the same server end point. adding one client only needs to get their page to subscribe to the very same app we create. this avoid the operation nightmare of horde of facebook apps (one per client per environment i.e. dev/test/staging/prod) and is closer to what facebook app design preference (one app has ONE and only ONE callback URL while multiple pages can subscribe to this one app with its own page token).

PS: I do have another feature might be useful to your case. It was about extracting the webserver creation logic out of the botkit code (facebook and slack) and allow configurable mount point per bot controller (instead of hardcoded /facebook/receive). This was done by overwriting setupWebserver() and createWebhookEndpoints() logic. The use case is that we have other light services running on the server via express (i.e. bot operational status end point and etc) I haven't shared that through a pull request yet, but plan to do that, after I got this one settled first.

ouadie-lahdioui:
To extract the webserver creation logic out of the botkit code and allow configurable mount point per bot controller take a look at [this one](https://github.com/howdyai/botkit-starter-facebook/blob/master/components/routes/incoming_webhooks.js)

facebook_bot.js
```diff
@@ -99,10 +100,18 @@ if(ops.lt === false && ops.ltsubdomain !== null) {
     process.exit();
 }
 
+var access_token = process.env.page_token;
+try {
+    var access_token = JSON.parse(access_token);
```
anonrig: It's not common to use JSON object as a Environment variable. Can you think of another way to implement this?

facebook_bot.js
```diff
@@ -99,10 +100,18 @@ if(ops.lt === false && ops.ltsubdomain !== null) {
     process.exit();
 }
 
+var access_token = process.env.page_token;
```
anonrig: Is it access_token or page_token, can you make this consistent?

facebook_bot.js
```diff
@@ -27,6 +27,7 @@ This bot demonstrates many of the core features of Botkit:
 
     app_secret=<MY APP SECRET> page_token=<MY PAGE TOKEN> verify_token=<MY_VERIFY_TOKEN> node facebook_bot.js [--lt [--ltsubdomain LOCALTUNNEL_SUBDOMAIN]]
 
+  Use page_token='{<"PAGE_A_ID":"PAGE_A_TOKEN","PAGE_B_ID":"PAGE_B_TOKEN"}' to enable multi tenant
```
anonrig: This is not a valid JSON object.

lib/Facebook.js
```diff
@@ -407,17 +409,18 @@ function Facebookbot(configuration) {
...
-            function(err, res, body) {
-                if (err) {
-                    facebook_botkit.log('Could not subscribe to page messages');
-                } else {
-                    facebook_botkit.debug('Successfully subscribed to Facebook events:', body);
-                    facebook_botkit.startTicking();
-                }
-            });
-
+        var access_tokens = (typeof (configuration.access_token) === 'string') ? [configuration.access_token] : configuration.access_token;
```
anonrig: `[configuration.access_token]` is an array, but on the other hand, in the beginning of the pull request, you converted access_token into an object. Is it an object or an array? It's getting pretty confusing atm.

lib/Facebook.js
```diff
@@ -407,17 +409,18 @@ function Facebookbot(configuration) {
...
+        Object.keys(access_tokens).forEach(function(page) {
```
anonrig: If it's an array, you shouldn't use Object.keys.

lib/Facebook.js
```diff
@@ -518,55 +521,64 @@ function Facebookbot(configuration) {
...
-            request.delete('https://' + api_host + '/v2.6/me/messenger_profile?access_token=' + configuration.access_token,
-                {form: message},
-                function(err, res, body) {
-                    if (err) {
-                        facebook_botkit.log('Could not configure messenger profile');
-                    } else {
-                        facebook_botkit.debug('Successfully configured messenger profile', message);
-                    }
-                });
+            var access_tokens = (typeof (configuration.access_token) === 'string') ? [configuration.access_token] : configuration.access_token;
```
anonrig: Same goes for here

lib/Facebook.js
```diff
@@ -518,55 +521,64 @@ function Facebookbot(configuration) {
...
         },
         postAPI: function(message) {
-            request.post('https://' + api_host + '/v2.6/me/messenger_profile?access_token=' + configuration.access_token,
-                {form: message},
-                function(err, res, body) {
-                    if (err) {
-                        facebook_botkit.log('Could not configure messenger profile');
-                    } else {
+            var access_tokens = (typeof (configuration.access_token) === 'string') ? [configuration.access_token] : configuration.access_token;
+            Object.keys(access_tokens).forEach(function(page) {
```
anonrig: Same goes for here.

lib/Facebook.js
```diff
@@ -578,35 +590,51 @@ function Facebookbot(configuration) {
...
-                        cb(err);
-                    } else {
-
-                        var results = null;
-                        try {
-                            results = JSON.parse(body);
-                        } catch (err) {
-                            facebook_botkit.log('ERROR in messenger code API call: Could not parse JSON', err, body);
+            var access_tokens = (typeof (configuration.access_token) === 'string') ? [configuration.access_token] : configuration.access_token;
+            Object.keys(access_tokens).forEach(function(page) {
```
anonrig: Same goes for here.

lib/Facebook.js
```diff
@@ -578,35 +590,51 @@ function Facebookbot(configuration) {
...
+                                }
                             }
                         }
-                    }
-                });
+                    });
+            });
+        },
+        deleteAPI: function(message) {
+            var access_tokens = (typeof (configuration.access_token) === 'string') ? [configuration.access_token] : configuration.access_token;
```
anonrig: Same goes for here.

alamusi:
@ouadie-lahdioui thanks for the reference to the code! this is similar to what i have done within the two functions, i.e. setupWebserver() and createWebhookEndpoints(). I believe the only difference the sample code with the built-in was:

- handle config.validate_requests
- call graph subscribed_apps api

same goes to slack, the difference would be:

- secureWebhookEndpoints

did i get this correct?

facebook_bot.js
alamusi: got it. thanks for spotting out the typo!

facebook_bot.js
alamusi: this was inherited from the previous sample code. I can certainly update it to keep the consistency.

facebook_bot.js
alamusi: you are right. it is indeed not the most common way. i was hoping to make the minimum changes possible and make it backward compatible... 

in our app, we saved this info in database... but for sample purpose, do you prefer to use a json file instead if passing in json as a env variable really bothers you?

lib/Facebook.js
alamusi: access_token can either be a string or an object. i used the [] notation to access the value by given the key as a variable, instead of dot notation, which works for known properties only.

lib/Facebook.js
alamusi: correct. i agree that it is best NOT to use Object.keys on an array. 

on the other hand, the interesting fact is that even when you have an array, Object.keys also works and extract only the defined indexes. (i.e. var a = []; a[123]="abc"; b[456]="def"; Object.keys(a); //this will return ["123","456"])

lib/Facebook.js
alamusi: access_tokens is an object with key as page_id and value as page_token

lib/Facebook.js
alamusi: access_tokens is an object with key as page_id and value as page_token

lib/Facebook.js
alamusi: access_tokens is an object with key as page_id and value as page_token

lib/Facebook.js
anonrig: When you use:

```javascript
var a = [] //this is an array
a[123] = 'abc'; //now you convert it to an array that has a length of 124. 
```

alamusi: you are right. i should have used { 0: configuration.access_token } instead of [configuration.access_token]. let me correct this right now across the board.

alamusi:
@anonrig is there anything else you would like me to update for this pull request?

peterswimm:
@alamusi there are a number of conflicts to be resolved, can you take a look against the current state of the facebook.js?

alamusi:
@peterswimm done - just synced with the latest master

metriqio:
Do we have an ETA on merge for this one ?

thaume:
Hello guys, do we have an ETA on merge for this one ? Thanks

benbrown:
This is an important issue that will be increasingly important as more and more apps for Facebook are developed.

That said, I want to make sure we have a pretty clear and standardized approach to multi-tenancy.  This already exists in the Slack connector, and if we are to broaden support, I'd like to make sure it is implemented in roughly the same way.

There may also be a need for a broader multi-tenancy refactor that would be applicable to all the connectors.

So the bottom line is, this PR can't be merged without a lot of really careful consideration. Multi-tenancy for Facebook is not officially on our roadmap right now, but can be added!

thaume:
@benbrown would it make sense to start an issue (RFC style) to discuss further integration of multi-tenancy for all services ? cc @alamusi

jonchurch:
@thaume Yeah that would be great
There are ways to do this that can be used now, my fork as a rough example above.

But having an issue to discuss this stuff would be a good idea.

MMayla:
@goodpic I liked your solution, could you show us an example of your implementation? (repo of a chatbot that can communicate with 2 different pages)

kikajanovcik:
Hi all,

I have started working on this - the requirement for my case is to have one FB app, but possibility to subscribe to multiple pages, thus find out the access token by page id in the facebook connector.

My solution passes all pages in config the same way as @alamusi suggested ({"page_a_id":"page_a_token", "page_b_id":"page_b_token"}).

For backwards compatibility, if config.pages is not set, I will simply return the access_token from config.

Let me know if you like this approach or would prefer if I did things differently. Thanks!

Issue #737, opened 2017-03-27
Convo.say not working with Event API


airlikidh2:
Slack related:
So I spent 4 hours figuring why my convo.say was not triggered while I was trying to transition from RTM to Events API.

If using the Event API only, you need to call:
`controller.startTicking();`

Could this be added in the doc?
https://github.com/howdyai/botkit/blob/master/readme-slack.md#events-api

peterswimm:
@airlikidh2 care to submit the PR?

Bunk:
Thanks for this, I was experiencing the exact same issue.

Went ahead and submitted a documentation PR (#802) for you.

lpedromachado:
I spent several hours too trying to make `convo.say()` work. It was only after I wrote an issue and I was typing its title that I saw this one. Now it works.

I see nowhere in the documentation either of `bot.startConversation()` or `convo.say()` an instruction to use `controller.startTicking()` first. Even the section about `startTicking()` doesn't say where to put it. I think the docs should be clearer about this.

benbrown:
@lpedromachado thanks for the feedback, i agree that startTicking is poorly documented.

PR #802, opened 2017-04-26
#737 Update Slack docs to include `startTicking` when using Events API

Bunk:
The code built into the Slack bot will call `controller.startTicking()` when calling `controller.startRTM()`; however, when using the Events API only `startTicking` will never get called and begin responding to conversations and threads.

There may be a better place to handle this via code, but a workaround is to manually call `startTicking` in this case.  The documentation could use an update to reflect this and prevent others from spending hours debugging why their conversations aren't working.


benbrown:
thanks for this @bunk!

PR #808, opened 2017-04-27
Change ping to an explicit message rather than ws protocol frame

aoberoi:
The current recommendation from Slack is to use a message to ping the websocket servers in addition to the protocol specific ping frames (the ws library can handle this all on its own and there's no need for code). Thus the ping code here was redundant since it was sending the same type of ping the server would have already been sending and the botkit client would have already been ponging. With this change, I also hope to resolve possible issues related to pings being an unmasked message and people getting abnormal terminations after that message was sent.

### DIFF
--- a/lib/Slackbot_worker.js
+++ b/lib/Slackbot_worker.js
@@ -186,7 +186,8 @@ module.exports = function(botkit, config) {
                         return;
                     }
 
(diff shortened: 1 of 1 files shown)


aoberoi:
this PR is redundant now, so i'll clean up house a bit :)

PR #813, opened 2017-04-28
Fix: message.im event name

neutraltone:
Fixed `message.im` event name in the readme as per documentation here: https://api.slack.com/events/message.im


peterswimm:
@neutraltone got catch!

PR #821, opened 2017-05-01
Code Improvement:

blasvicco:
Using inline if

### DIFF
--- a/lib/Slack_web_api.js
+++ b/lib/Slack_web_api.js
@@ -127,7 +127,7 @@ module.exports = function(bot, config) {
      * @param {function} cb A NodeJS-style callback
      */
     slack_api.callAPI = function(command, data, cb, multipart) {
-        data.token = data.token ? data.token : config.token;
+        data.token = data.token || config.token;
         bot.debug(command, data);
         postForm(slack_api.api_url + command, data, cb, multipart);
     };
@@ -140,9 +140,9 @@ module.exports = function(bot, config) {
      * @param {function} cb A NodeJS-style callback
      */
     slack_api.callAPIWithoutToken = function(command, data, cb) {
-        data.client_id = data.client_id ? data.client_id : bot.config.clientId;
-        data.client_secret = data.client_secret ? data.client_secret : bot.config.clientSecret;
-        data.redirect_uri = data.redirect_uri ? data.redirect_uri : bot.config.redirectUri;
+        data.client_id = data.client_id || bot.config.clientId;
+        data.client_secret = data.client_secret || bot.config.clientSecret;
+        data.redirect_uri = data.redirect_uri || bot.config.redirectUri;
         // DON'T log options: that could expose the client secret!
         postForm(slack_api.api_url + command, data, cb);
     };
@@ -160,7 +160,7 @@ module.exports = function(bot, config) {
         var currentGroup = slack_api;
 
         groups.forEach(function(nextGroupName) {
-            currentGroup[nextGroupName] = currentGroup[nextGroupName] ? currentGroup[nextGroupName] : {};
+            currentGroup[nextGroupName] = currentGroup[nextGroupName] || {};
             currentGroup = currentGroup[nextGroupName];
         });
 
--- a/lib/Slack_web_api.js
+++ b/lib/Slack_web_api.js
@@ -127,10 +127,7 @@ module.exports = function(bot, config) {
      * @param {function} cb A NodeJS-style callback
      */
     slack_api.callAPI = function(command, data, cb, multipart) {
-        if (!data.token) {
-            data.token = config.token;
-        }
-
+        data.token = data.token ? data.token : config.token;
         bot.debug(command, data);
         postForm(slack_api.api_url + command, data, cb, multipart);
     };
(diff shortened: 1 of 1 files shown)


jonchurch:
Thank you for your effort. 

I dont think that there are any peformance gains here, and personally find this to be less readable code.

blasvicco:
No problem dude. It is not a performance improvement. It is a personal suggestion to avoid unnecessary lines.

Cheers ;)

lib/Slack_web_api.js
```diff
@@ -127,10 +127,7 @@ module.exports = function(bot, config) {
      * @param {function} cb A NodeJS-style callback
      */
     slack_api.callAPI = function(command, data, cb, multipart) {
-        if (!data.token) {
-            data.token = config.token;
-        }
-
+        data.token = data.token ? data.token : config.token;
```
anonrig: `data.token = data.token || config.token` will make it more readable.

lib/Slack_web_api.js
```diff
@@ -143,16 +140,9 @@ module.exports = function(bot, config) {
...
-        if (!data.client_secret) {
-            data.client_secret = bot.config.clientSecret;
-        }
-        if (!data.redirect_uri) {
-            data.redirect_uri = bot.config.redirectUri;
-        }
-
+        data.client_id = data.client_id ? data.client_id : bot.config.clientId;
+        data.client_secret = data.client_secret ? data.client_secret : bot.config.clientSecret;
+        data.redirect_uri = data.redirect_uri ? data.redirect_uri : bot.config.redirectUri;
```
anonrig: Please use the `OR` operator for these operations

lib/Slack_web_api.js
```diff
@@ -170,9 +160,7 @@ module.exports = function(bot, config) {
         var currentGroup = slack_api;
 
         groups.forEach(function(nextGroupName) {
-            if (!currentGroup[nextGroupName]) {
-                currentGroup[nextGroupName] = {};
-            }
+            currentGroup[nextGroupName] = currentGroup[nextGroupName] ? currentGroup[nextGroupName] : {};
```
anonrig: Again `currentGroup[nextGroupName] = currentGroup[nextGroupName] || {}`

lib/Slack_web_api.js
```diff
@@ -195,11 +183,7 @@ module.exports = function(bot, config) {
 
     // specify that files get uploaded using multipart
     slack_api.files.upload = function(options, cb) {
-        if (options.file) {
-            slack_api.callAPI('files.upload', options, cb, true);
-        } else {
-            slack_api.callAPI('files.upload', options, cb, false);
-        }
+        slack_api.callAPI('files.upload', options, cb, !!options.file);
```
anonrig: 👍

lib/Slack_web_api.js
```diff
@@ -251,10 +235,7 @@ module.exports = function(bot, config) {
                     return cb(parseError);
                 }
 
-                if (json.ok) {
-                    return cb(null, json);
-                }
-                return cb(json.error, json);
+                return cb((json.ok ? null : json.error), json);
```
anonrig: 👍

lib/Slack_web_api.js
blasvicco: The problem with the || operator is that if the var is not defined it will throw an error.
Example:
testFunction(i) {
i = i || 10; //PERFECT
let t = t || 10; //ERROR because t is not defined
}

lib/Slack_web_api.js
blasvicco: Same as above

lib/Slack_web_api.js
blasvicco: Same as above

lib/Slack_web_api.js
anonrig: If data is defined on the same scope and in the same line, JS interpreter will throw an error. You're right. But the thing is `data` variable is defined before this line and therefore will not throw an error if data.token is undefined.

Example:

```javascript
> var data = {}
< undefined
> data.token = data.token || 3;
< 3
> data.token
< 3
```

blasvicco: Cool!!!
I will change it ;)

lib/Slack_web_api.js
```diff
@@ -170,9 +160,7 @@ module.exports = function(bot, config) {
         var currentGroup = slack_api;
 
         groups.forEach(function(nextGroupName) {
-            if (!currentGroup[nextGroupName]) {
-                currentGroup[nextGroupName] = {};
-            }
+            currentGroup[nextGroupName] = currentGroup[nextGroupName] || {};
             currentGroup = currentGroup[nextGroupName];
```
anonrig: Can you remove line 163 and replace this line with:

```javascript
currentGroup = currentGroup[nextGroupName] || {}
```

jonchurch:
Hey so why did this get merged in? There are only performance losses with ternary operators when replacing If/Else blocks in Javascript, and I don't find this to be more readable code.

@anonrig can you give some context to your choice here?

PR #822, opened 2017-05-01
Activate Facebook messenger profile API

ouadie-lahdioui:
Hello,

Enable a disabled section of Facebook messenger_profile examples.


ouadie-lahdioui:
This just illustrate some examples of how to use Facebook profile API.

If you prefer to let's the examples commented no problem !!

PR #824, opened 2017-05-02
🐛 trim space characters in events string

### LINKED ISSUE #807: Message Type string in #hears function
I've noticed this a few times, always forget to bring this up.

Spaces between the message type string in hears functions will register the first type, but not any others. Not sure that this is a bug, or just undocumented.

**The following will only trigger on direct_messages, but not ambient.**
`controller.hears('hi', 'direct_message, ambient', function(bot, message) { ... }`

**Without the space between direct_message and ambient they both trigger:**
`controller.hears('hi', 'direct_message,ambient', function(bot, message) { ... }`

Another valid way to do this is in an array (which the docs mention):
`controller.hears('hi', ['direct_message', 'ambient'], function(bot, message) { ... }`

**Suggested changes:** accept spaces in the string, or leave it how it is but give an error if spaces are used. What do you think @peterswimm @benbrown?

This has tripped me up quite a few times, and I'm sure it happens to others as well. Can't recall seeing this info out there in another issue, so here you go!

jonchurch:
Trims whitespace from the events string passed to hears function in corebot. Passing a string separated by commas and spaces would previously trigger only for the first event. This PR allows comma plus space separated strings of event names to register the events for each.

Previously, given `'ambient, direct_message'`, it would split the string into an array that looked like this: `['ambient', ' direct_message']` notice the extra space in direct_message string. It would then register a listener to that event, with the space included, which could never be triggered by botkit.

I'm afraid this might be breaking, or introduce unintended results for people. However, it still will only trigger events they have explicitly stated, but developers might get surprised by their inert code running where it was not before. 

Closes #807

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -1007,7 +1007,7 @@ function Botkit(configuration) {
         }
 
         if (typeof(events) == 'string') {
-            events = events.replace(/ /g, '').split(/\,/g);
+            events = events.split(/\,/g).map(function(str) { return str.trim() })
         }
 
         for (var e = 0; e < events.length; e++) {
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -1007,7 +1007,7 @@ function Botkit(configuration) {
         }
 
         if (typeof(events) == 'string') {
-            events = events.split(/\,/g);
+            events = events.replace(/ /g, '').split(/\,/g);
         }
 
         for (var e = 0; e < events.length; e++) {


lib/CoreBot.js
```diff
@@ -1007,7 +1007,7 @@ function Botkit(configuration) {
         }
 
         if (typeof(events) == 'string') {
-            events = events.split(/\,/g);
+            events = events.replace(/ /g, '').split(/\,/g);
```
anonrig: This update will inhibit the use of multiple keyword events. For example, if at some point we want to use "super channel" instead of "super_channel", this would be a problem. I'm suggesting to use `.trim()` for every splitted keyword.

What do you think?

anonrig: Ex:

```javascript
events = events.split(/\,/g).map(str => str.trim())
```

jonchurch: 👍

benbrown:
Thanks @jonchurch and @anonrig!

PR #826, opened 2017-05-03
declare "that" in scope

gswalden:
"that" is likely being declared on the global object, which may lead to leaks.


jonchurch:
👏Good👏eye👏my👏friend👏

benbrown:
Thank you everyone!

PR #838, opened 2017-05-09
send a 200 response when receiving a slash command

fletchrichman:
sends an empty status 200 response when a slash command is received so that slack doesn't freak out


fletchrichman:
via Ben brown: 

this actually will break the use of replyPrivate and other slash command specifics that send the response via the http response payload.

There is a function called replyAcknowledge that does this, and should be called in your slash command handler

PR #841, opened 2017-05-12
Introduction of Platform Provisioning Documents

peterswimm:
We've put off having verbose documentation for platforms long enough, here is a crack at supporting detailed directions on provisioning bots for the platforms that Botkit supports. More on the way.


anonrig:
Can you add this to .npmignore file? Since we don't want to install all of these images when we do `npm i botkit`.

peterswimm:
@anonrig you could argue that a npm installer NEEDS these docs too. Do you propose we should opt out of everything BUT the readme.md, and have a TOC to the other docs only in that for NPM installers?

colestrode:
I think @anonrig's makes a good point: this is quite a bit to download. I also think the main usefulness of including these images in an npm package would be offline access. But if I'm offline, I'm not going to be able to provision a bot 😄 

Also a large number (easily over 90%) of my installs are on CI servers, so this would be an increase in build time for no real gain.

anonrig:
No body will ever read the docs when they install it from NPM. Again it's your choice, but this will increase the package size and do no good to the developer. @colestrode @peterswimm

peterswimm:
Thanks for the feedback, my gut is to leave them in the project, but we will take your feedback under advisement.

benbrown:
I am going to add an .npmignore file and exclude these images, but leave the docs.

This is my own personal bias, but I *do* occasionally open up node_modules to find the docs.

PR #844, opened 2017-05-15
Moves examples under the example directory

colestrode:
Since we already have an examples directory, I think it would clean things up to move all example bots under the `/examples` directory. This helps identify where to look for help vs where to look for the library code.

To that end, this PR:
* Moves the original contents of `/examples` to `/examples/slack` since these were Slack specific examples
* Moves the top level `<platform>_bot.js` examples under `/examples`
* Updates all readme, documentation, and changelog links to point to the correct file.

I left `slack_bot.js` in the `/examples` directory rather than moving it into `examples/slack` for consistency. The `examples/slack` has detailed examples of Slack specific features, whereas `slack_bot.js` functions as a base example of using Slack and Botkit

Slack discussion here: https://dev4slack.slack.com/archives/C57S6LLS0/p1494874728641413


colestrode:
I also removed the `.vscode` directory and added that to `.gitignore`. These are user specific settings and shouldn't be in version control (I'd hate for someone to get tripped up because someone else's settings creeped in 😄 )

colestrode:
Just curious if this is going to be merged in soon? It's been approved for about a week and a half 😄

benbrown:
@colestrode merged! Thanks dude.

PR #854, opened 2017-05-20
Fix- Delete functions parameter fixed in simple_storage.js 

mehamasum:
fixed: 1st parameter of all 3 delete functions were mistaken


benbrown:
Thanks!

Issue #855, opened 2017-05-21
Unresolved callback on FindAppropiateTeam


alsuga:
Don't exists else for this [if] (https://github.com/howdyai/botkit/blob/master/lib/SlackBot.js#L154 ) and callback is not called

alecl:
I found fixed this in #887 before realizing it was recorded as an issue. I included the team_id there which is an added help for debugging.

jonchurch:
@peterswimm 🚫

PR #856, opened 2017-05-21
fixing unresolved callback on SlackBot

alsuga:
fixing #855


tmundt:
Should there be a PR?

benbrown:
This has been handled via a slightly different method. A fix is in master and will be included in the next release!

PR #862, opened 2017-05-26
customizable webhook endpoint path

birdietiger:
Cisco Spark webhook endpoint path was hardcoded as /ciscospark/receive. Ideally, it should be based on the public_address env variable.


benbrown:
The goal of the built-in webhook functions is to provide VERY SIMPLE and UNIFORM endpoints across platforms.

If you want to customize this path, we suggest following the example set forth in the botkit-starter-ciscospark project and building a custom endpoint along with `controller.handleWebhookPayload()`.

Issue #233, opened 2016-05-19
filedata structured message


vs1682:
I'm not able to find a way to send image files in structured messages.
After looking at the following code, I guess botkit has not implemented that part.
` bot.send = function(message, cb) {

            var facebook_message = {
                recipient: {},
                message: {}
            };

            if (typeof(message.channel) == 'string' && message.channel.match(/\+\d+\(\d\d\d\)\d\d\d\-\d\d\d\d/)) {
                facebook_message.recipient.phone_number = message.channel;
            } else {
                facebook_message.recipient.id = message.channel;
            }

            if (message.text) {
                facebook_message.message.text = message.text;
            }

            if (message.attachment) {
                facebook_message.message.attachment = message.attachment;
            }

            request.post('https://graph.facebook.com/me/messages?access_token=' + configuration.access_token,
                function(err, res, body) {
                    if (err) {
                        botkit.debug('WEBHOOK ERROR', err);
                        return cb && cb(err);
                    }

                    try {

                        var json = JSON.parse(body);

                    } catch (err) {

                        botkit.debug('JSON Parse error: ', err);
                        return cb && cb(err);

                    }

                    if (json.error) {
                        botkit.debug('API ERROR', json.error);
                        return cb && cb(json.error.message);
                    }

                    botkit.debug('WEBHOOK SUCCESS', body);
                    cb && cb(null, body);
                }).form(facebook_message);
        };`

and a look at facebook's Send API
`curl  \
  -F recipient='{"id":"USER_ID"}' \
  -F message='{"attachment":{"type":"image", "payload":{}}}' \
  -F filedata=@/tmp/testpng.png \
  "https://graph.facebook.com/v2.6/me/messages?access_token=PAGE_ACCESS_TOKEN"`

If I'm right botkit has not implemented that filedata part.

agamrafaeli:
Started working on this. Having a little trouble with the FB way of sending files. Would appreciate any help with my small technical problem:
http://stackoverflow.com/questions/40811931/how-can-i-send-a-file-to-facebook-send-api-using-npms-request-without-uploading

Ferrari:
Any update or plan on this feature?

PR #870, opened 2017-06-03
Support fb send upload

Ferrari:
Solved #233 

Use `filedata` to assign upload file/image describe on Content Types page in FB messenger doc. It will do a basic file validate (fs.existsSync) then stream the file using form-data.

e.g., upload png file
```
bot.reply(message, {
  attachment: {
    type: 'image',
    payload: {}
  },
  filedata: UPLOAD_FILE_PATH
})
```

### DIFF
--- a/lib/Facebook.js
+++ b/lib/Facebook.js
@@ -141,7 +141,7 @@ function Facebookbot(configuration) {
                 if (key !== 'filedata' && typeof facebook_message[key] === 'object') {
                   prepareFormdata[key] = JSON.stringify(facebook_message[key]);
                 } else if (key === 'filedata') {
-                  if (!fs.existSync(facebook_message[key])) return cb(new Error('filedata not exist'));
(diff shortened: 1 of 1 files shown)


lib/Facebook.js
```diff
@@ -129,31 +134,51 @@ function Facebookbot(configuration) {
...
 
+                  prepareFormdata[key] = fs.createReadStream(facebook_message[key]);
+                } else {
+                  prepareFormdata[key] = facebook_message[key];
+                }
+              }
+              requestObj = {
+                  method: 'POST',
+                  formData: prepareFormdata,
+                  uri: 'https://' + api_host + '/v2.6/me/messages'
```
jonchurch: Haven't used the upload api myself, but the [docs](https://developers.facebook.com/docs/messenger-platform/send-api-reference/attachment-upload) say the url for uploads is:
`v2.6/me/message_attachments`

PR #872, opened 2017-06-06
Create a web channel for Botkit

benbrown:
This pull request adds a web socket server, which can host direct conversations between a web or native client!

This is still UNDER DEVELOPMENT and we are officially requesting feedback from the community on this functionality!


lib/Sockets.js
```diff
@@ -0,0 +1,209 @@
...
+         // search through all the convos, if a bot matches, update its ws
+         var bot = controller.spawn();
+         bot.ws = ws;
+         bot.connected = true;
+
+         ws.on('message', function incoming(message) {
+
+           console.log('received: %s', message);
+
+           var message = JSON.parse(message);
```
jonchurch: Is try catch needed here?

jonchurch:
Stoked to see this happening! 

Is this intended to be compatible with existing messaging services/projects, or is there a client on the roadmap along with this? To deal with supporting buttons and displaying rich content.

lib/Sockets.js
munawwarhussain: needed

Naktibalda:
@benbrown hi, when are you going to merge this pull request?

lib/Sockets.js
```diff
@@ -0,0 +1,227 @@
...
+               } catch(err) {
+                   console.error('ERROR SENDING', err);
+               }
+           } else {
+               console.log('not connected. wait to resend');
+               setTimeout(function() {
+                   bot.send(message, cb);
+               }, 3000);
+           }
+       }
```
Naktibalda: Is there an interface that all chat integrations have to implement?
Facebook and Slackbot_worker have `startTyping`, `stopTyping`, `replyWithTyping` and `reply` methods.
Sockets `bot.send` method behaves like [replyWithTyping](https://github.com/howdyai/botkit/blob/07b7833869cf95d6aa714ba9b946e5633314b7dd/lib/Facebook.js#L173), but it is more primitive, because it types for 1 second when other libraries calculate how long the typing should last.

Naktibalda:
Outstanding pull requests to this branch: #973, #988 

Suggestion: I think that Sockets.js file should be renamed to SocketBot.js for consistency with other files.

Naktibalda:
@benbrown when are you going to merge this pull request?

peterswimm:
@Naktibalda the full release treatment requires some work and planning. Once we are feature complete on this, we can then start working on the general release with our usual accompaniment of documentation, examples, starter kits, and helper content.

In the meantime, this branch should be fully usable!

lib/Sockets.js
```diff
@@ -0,0 +1,269 @@
...
+            if (bot.connected) {
+                bot.ws.send(JSON.stringify({
+                    type: 'typing',
+                }));
+            }
+        };
+
+        bot.typingDelay = function(message) {
+
+            return new Promise(function(resolve, reject) {
```
Naktibalda: There is no `require("promise")` at the top  of this file, so this code is using Ecmascript6 Promise object.

I've never used it, so I don't know if it is important, but botkit has dependency on `promise` library: https://github.com/howdyai/botkit/blob/dev/webchannel/package.json#L24 which is used in https://github.com/howdyai/botkit/blob/dev/webchannel/lib/Studio.js#L2 and I think that it would be the best to use the same promise library in Sockets.js

PR #877, opened 2017-06-09
Fix ConsoleBot example

### LINKED ISSUE #714: Consolebot - Convo not working
Hi guys,

I'm trying to make run a simple convo from your examples with the `lib/consolebot` but it fails to reply to me on the `convo.say`

The `convo.ask` after the `startConversation` runs well, the callback is well fired, but nothing happens on the `say` triggering the next reply. 

Any idea ? 
Alex

aaronschachter:
* Fixes #714 by adding code courtesy of @alxpereira 's fix to define `bot.send`
* Updates the `uptime` event hander to properly respond

cc @peterswimm

### DIFF
--- a/examples/console_bot.js
+++ b/examples/console_bot.js
@@ -196,14 +196,12 @@ controller.hears(['shutdown'], 'message_received', function(bot, message) {
 
 
 controller.hears(['uptime', 'identify yourself', 'who are you', 'what is your name'],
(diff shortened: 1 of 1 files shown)


benbrown:
Thank you for this! 
I made these same changes, but in a slightly different place. They're in master now,  and will be included in the next release of Botkit!

PR #887, opened 2017-06-18
Add missing error callback from team not found

alecl:
Not calling this error callback leads to a weird hard to debug state if you've changed your data store and forgot to re-auth the team for example.


alecl:
Realized it's somewhat a duplicate of #856 but I include the team_id in this error which is an improvement.

PR #891, opened 2017-06-19
adds use strict directive to test files that use ES6

colestrode:
I think Node 4 requires the `use strict` directive to use `let`, `const`, and some other keywords. Under Node4 you will probably see this error:

```
SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode

      at ScriptTransformer._transformAndBuildScript (node_modules/jest-cli/node_modules/jest-runtime/build/ScriptTransformer.js:289:17)
      at emitTwo (events.js:87:13)
```

This PR adds the `use strict` directive to fix these types of errors.

### DIFF
--- a/__test__/lib/Slack_web_api.test.js
+++ b/__test__/lib/Slack_web_api.test.js
@@ -174,7 +174,20 @@ describe('postForm', () => {
             expect(cb).toHaveBeenCalledWith(error);
         });
 
-        test(`${methodName}: handles non 200 response code`, () => {
+        test(`${methodName}: handles 429 response code`, () => {
+            mockRequest.post.mockImplementation((params, callback) => {
+                callback(null, { statusCode: 429 }, null);
+            });
+
+            method('some.action', 'data', cb);
+
+            expect(mockRequest.post).toHaveBeenCalledTimes(1);
+            expect(cb).toHaveBeenCalledTimes(1);
+            const firstArg = cb.mock.calls[0][0];
+            expect(firstArg.message).toBe('Rate limit exceeded');
+        });
+
+        test(`${methodName}: handles other response codes`, () => {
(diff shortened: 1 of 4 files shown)


__test__/lib/Slack_web_api.test.js
```diff
@@ -172,7 +174,20 @@ describe('postForm', () => {
             expect(cb).toHaveBeenCalledWith(error);
         });
 
-        test(`${methodName}: handles non 200 response code`, () => {
+        test(`${methodName}: handles 429 response code`, () => {
```
colestrode: this branch was uncovered by tests, probably from a new PR

__test__/lib/Slack_web_api.test.js
```diff
@@ -253,41 +268,41 @@ describe('api methods', () => {
     describe('special cases', () => {
 
         test('chat.postMessage stringifies attachments', () => {
-            instance.chat.postMessage({attachments: []}, cb);
-            expect(instance.callAPI).toHaveBeenCalledWith('chat.postMessage', {attachments: '[]'}, cb);
+            instance.chat.postMessage({ attachments: [] }, cb);
```
colestrode: ew, sorry about all the whitespace changes. I think those are due to fixing linting errors.

lib/Slack_web_api.js
```diff
@@ -227,10 +227,11 @@ module.exports = function(bot, config) {
 
         request.post(params, function(error, response, body) {
```
colestrode: I refactored this method a bit to handle the case of an `error` and "bad" status codes separately. This rate limiting `if` would have thrown an unhandled exception if `request.post` called the callback with an error (thanks for catching that unit tests!)

PR #918, opened 2017-07-04
CiscoSpark version : inform when message come from not authorized dom…

MathRobin:
CiscoSpark version : inform when message come from not authorized domain or org

Useful for debug


MathRobin:
Thanks for your merge^^

benbrown:
Thanks for your contributions, @MathRobin!

PR #933, opened 2017-07-20
[Sockets] send calculates typing length

Naktibalda:
Copied from Facebook [replyWithTyping ](https://github.com/howdyai/botkit/blob/master/lib/Facebook.js#L173-L194) function.

#872


benbrown:
Merging!

PR #934, opened 2017-07-22
[CoreBot] Hears Control Flow - Run Multiple Hears Matches

### LINKED ISSUE #930: If a controller.hears fires, no other matching controller.hears will fire. Looking for ways around this.
Thought it might help to first explain the use-case here: I'm using controller.hears to look for references to various systems we use together. 

One module might be related to our development system, so might look for IDs matching BUG12345678, and will query the dev system for related info when that's seen. That module is dedicated to all listeners and functions related to the dev system, and any actions that might need to be taken with that system.

Another module might be related to our support case system, so might look for IDs matching Case1234456 and will query the support system to bring up a snapshot of related info, similar to the dev system module. The systems are completely different, so it's cleaner to have a dedicated module per system.

I'm using regex with controller.hears in each system's module within my slack bot to look for these system specific references. 

### Observation:
What I've noticed is that if more than one matching controller.hears exists, only one of them will actually fire. Not sure in what order. For example, if this kind of message comes up:

> For case 13245667 I'm not sure whether it fits to BUG123456 or BUG4958343, can anyone clarify?

### Desired result:
Right now, the case reference will be seen, and the slack bot will only trigger the controller.hears for the case ref. What I'm hoping to do is that any matching controller.hears trigger will fire, even if there's more than one. So the bot would:

- Run logic case 13245667.
- Run logic on BUG123456.
- Run logic on BUG4958343.

Anyone then getting involved in that discussion has immediate snapshots on all 3 matching references, without needing further action. 

### Questions:

- Can this be achieved with botkit right now? 
- Is there perhaps a flag preventing other controller.hears from firing on a single message, if a matching controller.hears already fired?

I guess the easiest way to do this right now is making a global controller.hears that reacts to every event, and then asking submodules if they want to react to that message.text, but that means a redesign contradicting the lovely modular capabilities of controller.hears only firing against matching messages via regex.

jonchurch:
This allows for returning early from a matched trigger, and continuing to the next (if any) matches.

This can be used to respond to a single input with multiple hears triggers, or allows you to ditch executing the current handler and either continue to your fallback handler, or another handler if present.

Ended up needing this for a project where bots are dynamically added to a server, and a bot can be either active or disabled, so if it's disabled I don't want to execute it's hears triggers. This allows me to add new bot skills to the server whenever, and enabled/disable the bots based on config details passed to them, so when a trigger fires they aren't configured for or have disabled, it will just keep executing and give the fallback.

closes #930

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -142,6 +142,43 @@ controller.hears(['keyword','^pattern$'],['message_received'],function(bot,messa
 });
 ```
 
+#### Control Flow
+
+By default, only the first succesfully matched hears handler is executed. Return `true` in your handler to stop executing the current handler and move on to the next hears match.
+
+Returning nothing, or `false`, will ensure no more handlers run, which is the default behavior.
+
+
+```javascript
+// sending 'bonjour' will respond in french if speaks.french is true
+// otherwise, the bot will respond with our catchall handler
+controller.hears(['^bonjour$'], 'direct_message', function(bot, message) {
+	if (speaks.french) {
+		bot.reply(message, 'Bonjour!')
+	} else {
+	// return true to move on to the next handler
+		return true
+	}
+})
+
+// sending hello will only respond with hello
+controller.hears(['^hello$'], 'direct_message', function(bot, message) {
+	bot.reply(message, 'Hello there!')
+})
+
+// sending howdy will give no response at all
+controller.hears(['^howdy$'], 'direct_message', function(bot, message) {
+	return false
+	bot.reply(message, 'nobody will ever see this')
+})
+
+controller.hears(['(.*)'], 'direct_message', function(bot, message){
+	bot.reply(message, 'I dont understand. (I respond to everything I hear)')
+})
+
+```
+
+
 When using the built in regular expression matching, the results of the expression will be stored in the `message.match` field and will match the expected output of normal Javascript `string.match(/pattern/i)`. For example:
 
 ```javascript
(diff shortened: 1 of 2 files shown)


benbrown:
This is an interesting solution to this problem.

One concern here: using a return value from the handler won't work if, inside that handler, you do anything asynchronous.

The counter-point may well be: regardless of what the handler does, it should tell Botkit to continue processing other handlers. 

I just want to make sure that this change, which would be at a VERY central part of Botkit, is done responsibly and in a way that supports most use cases.

tiagosiebler:
@benbrown I did face that concern while implementing this in my current botkit implementation. The general scenarios where this is leveraged is: if this trigger fires, regardless of what happens inside it, would I want to fire other potential triggers if they also match? Or should this be the only trigger that fires, regardless of the outcome?

For me that automatically means that in the rare scenarios where I don't want more than one trigger to fire, I don't need an async call to determine that and return false. For the rest, I can return true after starting any async requests, with no interest in how these requests complete (in the context of letting other triggers fire). I think that fits with your counter-point as well. 

Another thought is whether this change is different to how botkit works now by default, to which the answer is no. Unless you explicitly return true, botkit behaves like it did before this change happened. I've seen behave as it did before the PR, regardless whether I return false or return nothing at all. That's my experience with this solution so far. Curious if there are different ways this can/should be implemented.

benbrown:
thanks for your comment @tiagosiebler, very helpful.

I will try to spend time on this and come to some resolution soon.

willypt:
@benbrown Is this PR still alive and slated for next release?

tiagosiebler:
I'm still interested in an official implementation of this. There's value in matching multiple conditions with different handlers.

benbrown:
We are not going to add this feature.  If you want to fire multiple handlers for incoming messages, use the `controller.on()` method instead of `controller.hears()`

PR #952, opened 2017-08-10
Add error management in CoreBot

xcambar:
tl;dr; This PR adds error management into BotKit.

Motivation: We are working on a bot for which we need to limit interactions with the bot to a certain list of users. Hence, we need to skip processing if the user is not in the list of "authorised" users, and the code related to this feature belongs in a middleware.

Implementation: I leveraged `ware` and its already available management of JS exceptions (see https://github.com/segmentio/ware/blob/master/test/index.js). The idea is as follows: any `Error` thrown or passed to `next()` stops the execution of the middleware, and the (BotKit) callback will then interrupt further processing and yield to the newly introduced `error` middleware.

I'll be happy to provide you with more details if required, but I'd prefer to wait for your initial comments instead of unnecessarily detailing :)

### DIFF
--- a/__test__/lib/CoreBot.test.js
+++ b/__test__/lib/CoreBot.test.js
@@ -1,0 +1,65 @@
+var CoreBot = require('../../lib/CoreBot');
+
+describe('CoreBot', function() {
+    test('it should use the error middleware when an exception is thrown in a middleware', function() {
+        var bot = CoreBot({});
+        var errorHandler = jest.fn();
+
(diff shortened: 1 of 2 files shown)


benbrown:
Good job @xcambar! This is actually being implemented in the dev/pipeline branch too! I'll make sure your implementation and ours line up!

xcambar:
Thanks! Looking forward to seeing it available!

xcambar:
Can I have an indication as to why the PR has been closed?

PR #957, opened 2017-08-15
Mutate outgoing messages to include user and team ids and timestamps

thepavangollapalli:


benbrown:
@thepavangollapalli We are implementing a change very similar to this in the `dev/pipeline` branch.  Outgoing messages will have a .to field which contains the user id.

benbrown:
This was included in 0.6!

PR #958, opened 2017-08-15
[Slack] 👻 Add Ephemeral Message Support

jonchurch:
First pass at this for everyone hungry to jump on the new features! 
### **TLDR; We have bot.whisper now!!**

This adds two new functions to Botkit to make use of the new [`chat.postEphemeral`](https://api.slack.com/methods/chat.postEphemeral) Web API method.

### **bot.whisper()**:
Works just like bot.reply, takes a source message and clones the channel and user (user is required for this api method) and sends it. We also set `as_user` to true to show the messages are coming from the bot user (or whoevers token you're making calls with). This can be overridden by setting `user: false` on the outgoing message body, or using the second new method:

### **bot.sendEphemeral()**:
Works like bot.send, takes a botkit message with a channel and user (again, user is required for this method by slack) and sends a message, unprompted by the user.

## Ephemeral in Conversations
~Currently there is shakey support of interactive messages~, **slack is not sending the original message in the trigger payload, which is normally sent along with interactive message payloads.** ~replyInteractive to ephemeral interactive messages doesnt update them either. I asked in the slack api channel but havent heard anything about this yet.~  

~That unfortunately makes ephemeral messages, at this time, not as feature rich as their standard counterparts, and doing complex user flows with updating their choices isnt possible currently in the ephemeral realm 👻~

Got a response from @episod! Interactive messages do work with ephem. However, slack is not passing the original message object along with the payload. To respond, you must construct a new message. I've experienced issues where my updated message is attributed to the App and not the bot. Not sure what I'm doing wrong still, passing `as_user: true` doesnt seem to work.

**To send an ephemeral message as part of a normal conversation, add `ephemeral: true` to the message body. See the updated docs for an example**

### DIFF
--- a/lib/Slackbot_worker.js
+++ b/lib/Slackbot_worker.js
@@ -4,7 +4,6 @@ var slackWebApi = require(__dirname + '/Slack_web_api.js');
 var HttpsProxyAgent = require('https-proxy-agent');
 var Back = require('back');
 
-
 module.exports = function(botkit, config) {
     var bot = {
         type: 'slack',
@@ -338,6 +337,10 @@ module.exports = function(botkit, config) {
     };
 
     bot.send = function(message, cb) {
+		if (message.ephemeral) {
+			bot.sendEphemeral(message, cb)
+			return
+		}
         botkit.debug('SAY', message);
 
         /**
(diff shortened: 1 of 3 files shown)


lib/Slackbot_worker.js
```diff
@@ -338,6 +337,10 @@ module.exports = function(botkit, config) {
     };
 
     bot.send = function(message, cb) {
+		if (message.ephemeral) {
```
jonchurch: I chose to retain a single send function, and just check if the message should be ephemeral or not

Using send like this also preserves access to these messages from send middlewares!

lib/Slackbot_worker.js
```diff
@@ -413,6 +416,39 @@ module.exports = function(botkit, config) {
...
+            channel: message.channel,
+            text: message.text || null,
+            user: message.user,
+            as_user: message.as_user || false,
+            parse: message.parse || null,
+            link_names: message.link_names || null,
+            attachments: message.attachments ?
+                JSON.stringify(message.attachments) : null,
+        };
+		console.log({message})
```
k001: Why not use `botkit.debug` ?

lib/Slackbot_worker.js
```diff
@@ -413,6 +416,39 @@ module.exports = function(botkit, config) {
...
+		console.log({slack_message})
```
k001: Same here, is better use `botkig.debug`

lib/Slackbot_worker.js
jonchurch: No good reason, ty for catching it

PR #972, opened 2017-08-31
[Typescript] Add actionParams attribute to Message interface

Naktibalda:
Sometimes adding action attribute is not enough and additional parameters has to be set.
This change allows to set additional parameters in actionParams field.

The alternative solution would be to allow adding any parameters to Message, by setting `[index: string]: any` at the top level.

### DIFF
--- a/lib/Botkit.d.ts
+++ b/lib/Botkit.d.ts
@@ -212,6 +212,9 @@ declare namespace botkit {
   }
   interface Message {
     action?: string;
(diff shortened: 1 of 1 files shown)


Naktibalda:
If anybody else needs this, it is possible to extends Message declaration by adding this code to your project:

```typescript
declare module 'botkit' {
    interface Message {
        actionParams ?: {
            [index: string]: any
        };
    }
}
```

Issue #967, opened 2017-08-24
Slack Verification Token for interactive messages and events


shishirsharma:
Slack specifies a Verification Token for interactive messages and events api. Botkit does not support this out of the box. 

This can't be handled via checking for `req.body.token` in `createWebhookEndpoints` callback, as events get triggered before the call of callback. Also `cb` does not have `req`.

shishirsharma:
If this check makes sense, I am willing to open a PR.

peterswimm:
duplicate of https://github.com/howdyai/botkit/issues/966

shishirsharma:
https://github.com/howdyai/botkit/pull/981

Issue #966, opened 2017-08-24
Interactive Messages SSL Check


shishirsharma:
I have noticed that every time I update my ngrok url in Slack Interactive Messages, I receive `500 Internal Server Error`.

Upon inspection, It seems that slack does a SSL Check on these urls, with body `{"ssl_check":"1","token":"AAAAAAAAAAAAAAAAAAAAAAAA"}`. We should handle this in Botkit.

I am willing to open a PR for this.

shishirsharma:
https://github.com/howdyai/botkit/pull/981

shishirsharma:
This is fixed and merged

PR #981, opened 2017-09-04
add Slack token verification

shishirsharma:
Slack provides a token to verify that requests are indeed coming from slack. 

I have put a optional test for this. If config has `clientVarificationToken` it will be tested against all the slash command and interactive requests. 

In addition to that when we set a new `ngrok` url in Slack, Slack does a ssl check and that breaks code. I have handled that too.

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -72,6 +72,7 @@ function Slackbot(configuration) {
         } else {
             slack_botkit.config.clientId = slack_app_config.clientId;
             slack_botkit.config.clientSecret = slack_app_config.clientSecret;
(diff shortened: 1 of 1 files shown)


shishirsharma:
https://github.com/howdyai/botkit/issues/967

shishirsharma:
https://github.com/howdyai/botkit/issues/966

shishirsharma:
I am bumping this up. Let me know if this approach is not favoured.

benbrown:
@shishirsharma thanks for the reminder we'll try to process this one soon.

benbrown:
Thanks!

PR #986, opened 2017-09-05
Use built-in crypto instead of package

EriksRemess:
crypto package is deprecated
more info https://www.npmjs.com/package/crypto


benbrown:
This change is only in the most recent versions of Node.

We need to investigate the impact before requiring newer Node version.

PR #991, opened 2017-09-05
Update to better syntax

teerapat2539:


peterswimm:
Can you explain what you are hoping to achieve with these changes?

PR #1001, opened 2017-09-06
add missing usergroups.* api endpoints to lib/Slack_web_api.js

digitalspecialists:
add missing usergroups.create, usergroups.disable, usergroups.enable, usergroups.list, usergroups.update, usergroups.users.list, usergroups.users.update to lib/Slack_web_api.js


digitalspecialists:
Would now also be useful to add the new conversations api methods, of which there are a bunch.
https://api.slack.com/docs/conversations-api

Issue #992, opened 2017-09-05
Facebook quick_reply should trigger an event


adrienbarreau:
When you click on a button with a postback, it triggers the event `facebook_postback`
When you click on a quick replies, which has also a postback, it triggers only the event `message_received`.

It would be nice to have an event dedicated to quick_replies or trigger the event `message_received`

afbayonac:
you can use a middleware
```  /* quick_replies */
  controller.middleware.normalize.use(function handleQuickReply (bot, message, next) {
    if (message.quick_reply) {
      message.text = message.quick_reply.payload
      message.payload = message.quick_reply.payload
      message.type = 'facebook_quick_reply'
    }
    next()
  })
```

PR #1013, opened 2017-09-10
add trigger facebook_quick_reply 

afbayonac:
#992


peterswimm:
There are some changes that are broken with the latest push, can you review the conflicting files and provide some background on the changes you are making for review?

lib/Facebook.js
```diff
@@ -315,6 +315,23 @@ function Facebookbot(configuration) {
...
 
+                    } else if (facebook_message.message && facebook_message.message.quick_reply) {
+
+                        // trigger BOTH a quick_reply event
+                        // and a normal message received event.
+                        // this allows developers to receive quick_reply as part of a conversation.
+                        var message = {
+                            text: facebook_message.message.quick_reply.payload,
+                            payload: facebook_message.message.quick_reply.payload,
+                            input: facebook_message.message.text,
```
ouadie-lahdioui: Why are you added an input entry to the message ?

afbayonac:
ready !!

benbrown:
This is interesting! 

I agree with @ouadie-lahdioui here - this would be a breaking change for a lot of old bots that expect quick replies to work like normal text.

This could be achieved with an _optional_ categorize middleware that looks for the quick_replies field and changes the message type.

Thoughts?

benbrown:
We are currently not going to change the way quick replies are handled.

PR #1023, opened 2017-09-13
Fix typo preventing Twilio Media Urls

jpowers:
Hey there,

Just found a small typo in the Twillio SMS lib.  You can work around it if you set both `mediaUrl` and `medaUrl` like:

```convo.say({text: 'Here\'s your URL', mediaUrl: url, medaUrl: url})```

But that makes me sad.


benbrown:
Thank you!!

PR #1044, opened 2017-09-19
Fixed typo for Facebook NLP disable api

shashank-saxena:


peterswimm:
Good catch!

benbrown:
thanks @shashank-saxena

PR #1057, opened 2017-09-22
Create test_your_bot_app.md

Raniazy:


Naktibalda:
I think that it would be more useful to provide a list of existing test frameworks.
I know https://github.com/codeforequity-at/testmybot and https://github.com/gratifyguy/botkit-mock

PR #1067, opened 2017-09-30
Updated http to https for some links

jimpallomeni:


peterswimm:
These redirect to HTTPS on the server side, but couldn't hurt to add!

PR #1070, opened 2017-10-01
Update team_slashcommand.js

NatJNP:
I change single comment to multilines comment. I think it look more beautiful code.


benbrown:
Sorry, we don't accept changes that only affect the formatting of code comments

Issue #1071, opened 2017-10-04
bot.whisper posts publicly to the channel instead of an ephemeral response in botkit 0.6.3


fletchrichman:
We are using the bot.whisper function in a public channel. It was working on 0.6.0, but now it is posting back to the whole channel. 

![image](https://user-images.githubusercontent.com/1915836/31155180-6c938ade-a869-11e7-97f6-5444f6b25026.png)

![image](https://user-images.githubusercontent.com/1915836/31155153-4d2dc24a-a869-11e7-8366-a4cada41fb91.png)

benbrown:
This should be fixed in 0.6.4!

PR #1073, opened 2017-10-04
[Slack] Fix whisper v0.6

jonchurch:
#1071 Says it all

This PR is not working, wanted to share my progress before going to sleep in my TZ

Currently ephemeral message info is not making it out of format middleware

I changed whisper to just set message type to `'ephemeral'` and set the `user` field on outgoing request to slack (required info for ephemeral messages).

But I ran into this error and I can't find where slack_message is referenced at all in format:
`info: An error occured in the format middleware: ReferenceError: slack_message is not defined`


benbrown:
This is fixed in 0.6.4! Thanks @jonchurch

PR #1098, opened 2017-10-19
Add bot.getMessageUser function to primary platforms

benbrown:
Adds a standardized mechanism for loading a standardized user profile for Slack, Teams, Cisco Spark and Facebook.

As a promise:
`bot.getMessageUser(message).then(function(profile) { ... });`

As a callback:
`bot.getMessageUser(message, function(err, profile {...});`

`profile` object will contain:

`{
username:
first_name:
last_name:
full_name:
gender:
timezone:
timezone_offset:
email:
}`

### DIFF
--- a/lib/CiscoSparkbot.js
+++ b/lib/CiscoSparkbot.js
@@ -481,7 +481,7 @@ function Sparkbot(configuration) {
               }
               resolve(profile);
 
-            }).catch(err) {
+            }).catch(function(err) {
               if (cb) {
                 cb(err);
               }
--- a/lib/CiscoSparkbot.js
+++ b/lib/CiscoSparkbot.js
@@ -459,14 +459,14 @@ function Sparkbot(configuration) {
 
         };
 
-        bot.getUser = function(user, cb) {
+        bot.getMessageUser = function(message, cb) {
             return new Promise(function(resolve, reject) {
 
-                controller.api.people.get(user).then(function(identity) {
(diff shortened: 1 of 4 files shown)


jonchurch:
To clarify, you want this to be a Botkit specific function, uniform experience between all platforms? I'm not sure of a case for such uniformity at this time, but as the controller gets unified maybe it'll be important. But this does leave out some things that are important platform specific uses, and will likely trip up people who have trouble finding things in the mountain of docs as it stands now.

**This feature would lead to having two different profile api calls, so maybe it can be made clear it is an internal Botkit function?**

In Facebook.js we don't currently have an API method to just get at the FB request for the [messenger user profile info](https://developers.facebook.com/docs/messenger-platform/identity/user-profile/), I think there will be confusion for people as to why this doesn't include some of the more important FB profile information.

Such as `profile_pic, last_ad_referral, is_payment_enabled, locale`

**Would be a good PR for someone to add at the user profile API call, ala [this middleware](https://github.com/mrbot-ai/botkit-middleware-fbuser/blob/master/src/botkit-middleware-fbuser.js) someone wrote.**

Profile pic is used as a hack to identify a user outside your Bot's scope (we get a page scoped ID, and it cant be used to link to someone using your other FB apps unless you compare the profile pic url)

Ad referral is getting more important as FB Messenger ads become more popular, helps with segmenting.

Locale is important too for internationalization

PR #1099, opened 2017-10-20
Include 'template' to possible FacebookAttachment types

rangelfinal:
A attachment may have one of the following types: "image", "audio", "video", "file" or "template", as defined here: https://developers.facebook.com/docs/messenger-platform/reference/send-api/#attachment


benbrown:
Thank you!

PR #1114, opened 2017-11-05
small addition, to make things clear

atherdon:
Before it was:
_"Interactive messages" is Slack's fancy way of saying "buttons." In order to enable buttons, under Request URL, add https://my-bot-url/slack/receive, then click save."_

Now: 
_Interactive messages" is Slack's fancy way of saying "buttons." In order to enable buttons, click on the "Interactive Components" tab and under Request URL, add https://my-bot-url/slack/receive, then click save._


peterswimm:
Thanks for this! We should go through all the docs and update little slack changes like this!

Issue #1146, opened 2017-11-27
Facebook sendApi errors are being swallowed


nishant-chaturvedi:
Currently Facebook bot in the botkit, swallows the `error` objects obtained as a result of `sendApi` call, and simply passes the error message (string) to the callbacks. This makes it impossible to check sendApi `error codes` and `error subcodes` and handle those differently based on the business use cases.

It would be very helpful to pass the entire error object as opposed to just the error message to the callbacks.

third1020:
Looks cool!

benbrown:
Merged and shipped!

PR #1147, opened 2017-11-27
pass error object to the callbacks for failed send Api calls

nishant-chaturvedi:
Fix for #1146


ouadie-lahdioui:
@nishant-chaturvedi LGTM

ouadie-lahdioui:
@nishant-chaturvedi LGTM 👍 

NB: It will be break changes for the users who wait for a simple string message and not an error object.

benbrown:
thank you!

PR #1151, opened 2017-11-29
[CoreBot] Don't suppress errors

Naktibalda:
I don't know if this change fixes any real issue, but not calling a callback when it is provided is obviously wrong.


benbrown:
Thank you!

PR #1153, opened 2017-11-29
[Sockets]  Suppress websocket exception in startTyping 

Naktibalda:
It will be followed by sending of a real message anyway and that exception will be handled properly.

### DIFF
--- a/lib/Sockets.js
+++ b/lib/Sockets.js
@@ -157,9 +157,11 @@ function SocketBot(configuration) {
 
         bot.startTyping = function() {
             if (bot.connected) {
-                bot.ws.send(JSON.stringify({
-                    type: 'typing',
-                }));
+                try {
+                  bot.ws.send(JSON.stringify({
+                    type: 'typing'
+                  }));
+                } catch (err) {}
             }
         };
 


Naktibalda:
Actually this is not good enough:


https://github.com/websockets/ws#error-handling-best-practices
```js
// Errors (both immediate and async write errors) can be detected in an optional
// callback. The callback is also the only way of being notified that data has
// actually been sent.
ws.send('something', function ack(error) {
  // If error is not defined, the send has been completed, otherwise the error
  // object will indicate what failed.
});

// Immediate errors can also be handled with `try...catch`, but **note** that
// since sends are inherently asynchronous, socket write failures will *not* be
// captured when this technique is used.
```

I will make another pull request with callback for handling errors.

Naktibalda:
The error I am getting is this one:
```
Error: socket hang up
    at createHangUpError (_http_client.js:254:15)
    at TLSSocket.socketCloseListener (_http_client.js:286:23)
    at emitOne (events.js:101:20)
    at TLSSocket.emit (events.js:188:7)
    at _handle.close (net.js:497:12)
    at TCP.done [as _onclose] (_tls_wrap.js:332:7)
```
It gets caught by bluebird promises, but it comes up in a wrong place.

Naktibalda:
I think that my error is originating from callback-less socket.write call in https://github.com/websockets/ws/blob/8ea3739d62fd7b2e467ec814e31c7f65d32da98b/lib/Sender.js#L374-L375

```js
  sendFrame (list, cb) {
    if (list.length === 2) {
      this._socket.write(list[0]);
      this._socket.write(list[1], cb);
    } else {
      this._socket.write(list[0], cb);
}
```

Naktibalda:
No, it wasn't a websocket issue.
This error originates from TLSWrap.close https://github.com/nodejs/node/blob/v6.12.0/lib/_tls_wrap.js?utf8=%E2%9C%93#L332

Naktibalda:
Blaming and debugging random packages isn't a particularly productive activity.
My error was actually a result of HTTP timeout of request made after sending reply.

PR #1154, opened 2017-11-30
Upgrade ws library

Naktibalda:
Breaking changes in [3.0.0](https://github.com/websockets/ws/releases/tag/3.0.0) were:
```
    Removed the upgradeReq property (#1099).
    Removed unnecessary events (#1100).
    Removed the flags argument from the 'message', 'ping', and 'pong'
    events (#1101).
    The permessage-deflate extension is now disabled by default on the server
    (#1107).
```

Only flag argument change affects botkit, but flags argument was unused.


Naktibalda:
@benbrown I made pull request to master branch because lib/Slackbot_worker.js uses websockets.

benbrown:
thank you!

PR #1164, opened 2017-12-07
Add delete method to in-memory store

Naktibalda:
Most storage methods have no delete method,
but [simple_storage(files)](https://github.com/howdyai/botkit/blob/master/lib/storage/simple_storage.js#L59),  [couchdb](https://github.com/barlock/botkit-storage-couchdb/blob/master/lib/storage.js#L60) and [beepboop](https://github.com/BeepBoopHQ/botkit-storage-beepboop/blob/master/src/storage.js#L84) storage modules have it, so I am adding it to memory store.

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -979,6 +979,10 @@ function Botkit(configuration) {
                     cb('No ID specified');
                 }
             },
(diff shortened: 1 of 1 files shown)


benbrown:
thanks @Naktibalda

PR #1170, opened 2017-12-08
Securing FB Graph API Requests

ouadie-lahdioui:
Hello,

Almost every Graph API call requires an access token. Malicious developers can steal access tokens and use them to send spam from your bot with malicious software on a person's computer or a man in the middle attack. 

To prevent that, Facebook recommend sending an app secret proof parameter to every API call. 

This PR add a new configuration attribute ```require_appsecret_proof``` to enable sending the sha256 proof each call Botkit makes.

````
var controller = Botkit.facebookbot({
    debug: true,
    log: true,
    access_token: process.env.page_token,
    verify_token: process.env.verify_token,
    app_secret: process.env.app_secret,
    require_appsecret_proof: true
});
````

Enjoy :v:

### DIFF
--- a/docs/readme-facebook.md
+++ b/docs/readme-facebook.md
@@ -21,6 +21,9 @@ Table of Contents
 * [Silent and No Notifications](#silent-and-no-notifications)
 * [Messenger code API](#messenger-code-api)
 * [Attachment upload API](#attachment-upload-api)
+* [Built-in NLP](#built-in-nlp)
+* [Message Tags](#message-tags)
+* [App Secret Proof](#app-secret-proof )
 * [Running Botkit with an Express server](#use-botkit-for-facebook-messenger-with-an-express-web-server)
 
 ## Getting Started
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/Facebook.js [hunk 1]
```diff
@@ -851,6 +909,11 @@ function Facebookbot(configuration) {
         return 'sha1=' + hmac.digest('hex');
     }
 
+    function getAppSecretProof(dataToHash, key) {
```
htaidirt: I'm probably missing the big picture, but is this method intended to be used somewhere else?

If not, I suggest to use more explicit parameter names, like:

```
function getAppSecretProof(access_token, app_secret) {
    var hmac = crypto.createHmac('sha256', app_secret);
    return hmac.update(access_token).digest('hex');
}
```

ouadie-lahdioui: totally agree 👍

docs/readme-facebook.md
```diff
@@ -570,6 +573,22 @@ var taggedMessage = {
 bot.reply(message, taggedMessage);
 ```
 
+## App Secret Proof 
+
+To improve security and prevent your bot against man in the middle attack, it's highly recommended to send an app secret proof : 
+
+```javascript
+var controller = Botkit.facebookbot({
+    debug: true,
```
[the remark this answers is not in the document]
ouadie-lahdioui: Done

ouadie-lahdioui:
Thanks @htaidirt for the review.

- Yes, it should be a lot of tests there, unfortunately not all functionalities are tested. fork the project and help adding some cool tests.
- In case the app secret is not provided Botkit will encrypt using an empty key.

PR #1171, opened 2017-12-08
Include messaging_type property in all Botkit message sends

ouadie-lahdioui:
Hello,

As of the release of Messenger Platform v2.2 including the messaging_type property in all message sends is become highly recommended.

Beginning May 7, 2018, this property will be required for all message sends !

For more information documentation is [here](https://developers.facebook.com/docs/messenger-platform/reference/send-api/)

Enjoy :sunglasses:


benbrown:
@ouadie-lahdioui Looks great!

Can you help us by adding some docs to the docs/readme-facebook.md file about when and how to use this new field? Like, when would a developer set the value?

ouadie-lahdioui:
Done. i added a messaging type section in the readme file.

PR #1180, opened 2017-12-11
Add FB Broadcast Messages API

ouadie-lahdioui:
Hello,

This PR add the new FB Broadcast API to send a message to many recipients.

Enjoy :alien:

### DIFF
--- a/lib/Facebook.js
+++ b/lib/Facebook.js
@@ -987,7 +987,29 @@ function Facebookbot(configuration) {
                     }
                 }
             });
-        }
+        },
+        get_labels_by_user: function(user, cb) {
+            var uri = 'https://' + api_host + '/' + api_version + '/' + user + '/custom_labels?access_token=' + configuration.access_token;
+
+            request({
+                method: 'GET',
+                json: true,
+                uri: uri
+            }, function(err, res, body) {
+                if (err) {
+                    facebook_botkit.log('Could not get labels associated with a user');
+                    cb(err);
+                } else {
+                    if (body.error) {
+                        facebook_botkit.log('ERROR in getting labels associated with a user call : ', body.error.message);
+                        cb(body.error);
+                    } else {
+                        facebook_botkit.log('Successfully get labels associated with a user');
+                        cb(null, body);
+                    }
+                }
+            });
+        },
     };
 
     facebook_botkit.api = {
--- a/lib/Facebook.js
+++ b/lib/Facebook.js
@@ -906,7 +906,34 @@ function Facebookbot(configuration) {
                     }
                 }
             });
-        }
+        },
+        create_label: function(name, cb) {
+            var uri = 'https://' + api_host + '/' + api_version + '/me/custom_labels?access_token=' + configuration.access_token;
+
+            var body = {
+                'name' : name
+            };
+
+            request({
+                method: 'POST',
+                json: true,
+                body: body,
+                uri: uri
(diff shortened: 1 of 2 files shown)


lib/Facebook.js
```diff
@@ -815,10 +816,274 @@ function Facebookbot(configuration) {
...
+                        facebook_botkit.log('ERROR in getting user associated labels API call : ', body.error.message);
+                        cb(body.error);
+                    } else {
+                        facebook_botkit.log('Successfully get user associated labels');
+                        cb(null, body);
+                    }
+                }
+            });
+        },
+        get_label_details: function(label_id, fields, cb) {
```
jonchurch: Currently the only available extra field for a label is `name`, which I think we can just add as the default to this request and save the additional argument to get_label_details

lib/Facebook.js
```diff
@@ -815,10 +816,274 @@ function Facebookbot(configuration) {
...
+                        facebook_botkit.log('ERROR in getting label details call : ', body.error.message);
+                        cb(body.error);
+                    } else {
+                        facebook_botkit.log('Successfully get label details');
+                        cb(null, body);
+                    }
+                }
+            });
+        },
+        get_all_label: function(fields, cb) {
```
jonchurch: Maybe get_all_label should be get_all_labels with an s? there is get_labels_by_user above

Same comment about the fields, I think we can just grab all fields by default since FB only lets us specify `name` as an additional field currently

lib/Facebook.js
```diff
@@ -815,10 +816,274 @@ function Facebookbot(configuration) {
...
+                        facebook_botkit.log('ERROR in creating a broadcast message call: ', body.error);
+                        cb(body.error);
+                    } else {
+                        facebook_botkit.debug('Successfully create a broadcast message', body);
+                        cb(null, body);
+                    }
+                }
+            });
+        },
+        send: function(message_creative_id, custom_label_id,  cb) {
```
jonchurch: Take a look at the sendBroadcast function I created in this [PR](https://github.com/howdyai/botkit/pull/1181/files#diff-c54b974993736486b57fb75303b19213R943) where we test if a `creative_message_id` string is passed, or a fresh message/array of messages. I prefer that personally, saves the user a callback level if they don't need to store the creative_message_id

lib/Facebook.js
ouadie-lahdioui: I put fields to prevents any futur break changes, Exactly, today, name is the only available field, bu FB can add some other ones.

lib/Facebook.js
ouadie-lahdioui: Nice catch 👍

ouadie-lahdioui:
Thank you @jonchurch for the review 👍

senorcodecat:
Hey, is this going to be accepted into the base soon? Any timeline?

MagSag:
Hey this is great; does it support adding multiple users with one request to a custom label? I am looking on how I can add [user1,user2,user3] directly a custom label.

ouadie-lahdioui:
Hi @MagSag, FB broadcast API doesn't support associating multiple users to a label ! to add more than one user to specific label, you should make more than one api call.

The same thing for deleting a broadcast message, it's not yet supported by [FB](https://developers.facebook.com/docs/messenger-platform/send-messages/broadcast-messages)

Enjoy

PR #1183, opened 2017-12-15
Add Facebook insights API

### LINKED ISSUE #1126: [Facebook] Platform v2.2: Page Insights API
With Facebook Messenger Platform v2.2 Page Insights API was introduced [(docs)](https://developers.facebook.com/docs/messenger-platform/reference/messaging-insights-api/)

This is an API feature which should be added similarly to how the other FB api's are implemented in [Facebook.js](https://github.com/howdyai/botkit/blob/66464ae92849873e9ca0272c97b89beb7bb5588d/lib/Facebook.js#L823)

Anyone interested? Good first contribution for somebody 👍 

Let me know if I can help you!

ouadie-lahdioui:
Hello,

This PR add the new FB insights API to the awesome Botkit.

Fix #1126 

Enjoy

### DIFF
--- a/examples/facebook_bot.js
+++ b/examples/facebook_bot.js
@@ -488,6 +488,15 @@ controller.hears(['send tagged message'], 'message_received', function (bot, mes
     bot.reply(message, taggedMessage);
 });
 
+controller.hears(['insights'], 'message_received', function (bot, message) {
+    controller.api.insights.get_insights(['page_messages_active_threads_unique', 'page_messages_blocked_conversations_unique'], null, null, function (err, body) {
+        if (err) {
+            bot.reply(message, 'Insights error');
+        } else {
+            bot.reply(message, JSON.stringify(body.data));
+        }
+    });
+});
 
 function formatUptime(uptime) {
     var unit = 'second';
--- a/lib/Facebook.js
+++ b/lib/Facebook.js
@@ -815,9 +815,46 @@ function Facebookbot(configuration) {
         });
     };
 
+    var insights_api = {
+        get_insights: function (metrics, since, until, cb) {
+            var uri = 'https://' + api_host + '/v2.8/me/insights?access_token=' + configuration.access_token;
 
+            if (metrics) {
+                if (typeof(metrics) == 'string') {
+                    uri += '&metric=' + metrics;
+                } else {
+                    uri += '&metric=' + metrics.join(',');
+                }
+            }
 
+            if (since) {
+                uri += '&since=' + since;
+            }
 
+            if (until) {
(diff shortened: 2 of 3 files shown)


lib/Facebook.js
```diff
@@ -815,9 +815,46 @@ function Facebookbot(configuration) {
...
+                uri += '&until=' + until;
+            }
+
+            request({
+                method: 'GET',
+                json: true,
+                uri: uri
+            }, function (err, res, body) {
+                if (err) {
+                    facebook_botkit.log('Could call Facebook Insights API');
```
Naktibalda: I think that this log message is wrong.
Also logging err.message helps with debugging.

lib/Facebook.js
```diff
@@ -815,9 +815,46 @@ function Facebookbot(configuration) {
...
+                    cb(err);
+                } else {
+                    if (body.error) {
+                        facebook_botkit.log('ERROR in Facebook Insights API call: ', body.error.message);
+                        cb(body.error);
```
Naktibalda: body.error is not an instance of Error, is it?
If it isn't, then it would probably be better to use `cb(new Error(body.error.message));`

examples/facebook_bot.js
```diff
@@ -488,6 +488,11 @@ controller.hears(['send tagged message'], 'message_received', function (bot, mes
     bot.reply(message, taggedMessage);
 });
 
+controller.hears(['insights'], 'message_received', function (bot, message) {
+    controller.api.insights.get_insights(['page_messages_active_threads_unique', 'page_messages_blocked_conversations_unique'], null, null, function (err, body) {
+        bot.reply(message, JSON.stringify(body.data));
```
Naktibalda: A little bit of error handling would be nice here, especially because many people use examples with little modification.
body can be undefined, if that's the case then accessing body.data would throw an error.

```js
if (err) {
    bot.reply(message, 'Insights error');
} else {
    bot.reply(message, JSON.stringify(body.data));
}
```

ouadie-lahdioui: It looks nice, done 👍

lib/Facebook.js
ouadie-lahdioui: Nice catch 👍

lib/Facebook.js
ouadie-lahdioui: No it's not, ````body.error```` is the error object returned by FB, it contains other useful data than message.

If we call the callback with an error object and juste the message, we couldn't use those useful data (code, ...).

ouadie-lahdioui:
@Naktibalda thanks for the review :thumbsup:

PR #1184, opened 2017-12-15
Mount Slack auth middleware on the proper route

miklosaubert:
The Slack authentication middleware is wrongly mounted at application level, meaning it will handle all requests to any routes defined after a call to `createWebhookEndpoints`. To prevent this, we must ensure that the middleware is mounted only on the `/slack/receive` route.


benbrown:
This is a duplicate of #1203 but thanks!!

Issue #1212, opened 2018-01-15
Facebook only sends first message in convo if require_delivery is true


jsalwen:
Setup

```js
controller = Botkit.facebookbot({
  ...
  require_delivery: true
});

controller.hears('convo-test', 'message_received', (bot, message) => {
  bot.startConversation(message, function(err, convo) {
    convo.say('Look');
    convo.say({
      "attachment": {
        "type": "image",
        "payload": {
          "url": "https://c1.staticflickr.com/1/34/122530930_6e16f1eb5c.jpg"
        }
      }
    });
    convo.say('That is a fat cat!');
  });
});
```

If I type `convo-test` in messenger, I only get:
```
Look
```

## Cause

The message is never set to `delivered`
On line 27: when checking for a `mid` match, `convo.sent[s].api_response.mid` is `undefined`
```
convo.sent[s].api_response && convo.sent[s].api_response.mid == mid
```
the correct property is `message_id`
```js
// convo.sent[s].api_response
{
  recipient_id: '1127397383997557',
  message_id: 'mid.$cAAZoIpjMd8BnLT1FD1g-_f8eDAwu'
}
```

https://github.com/howdyai/botkit/blob/5d237d5839f0bf3715d143e9fcb5f8bab4581aa9/lib/Facebook.js#L13-L37

I will submit a PR

jsalwen:
see Facebook doco on this:
https://developers.facebook.com/docs/messenger-platform/reference/send-api#response

jsalwen:
workaround until PR #1213 is merged

is just repeats the basic logic from above, but with the fix
```js
controller.on('message_delivered', function(bot, message) {
  // get list of mids in this message
  for (var m = 0; m < message.delivery.mids.length; m++) {
    var mid = message.delivery.mids[m];

    // loop through all active conversations this bot is having
    // and mark messages in conversations as delivered = true
    bot.findConversation(message, function(convo) {
      if (convo) {
        for (var s = 0; s < convo.sent.length; s++) {
          if (convo.sent[s].api_response && convo.sent[s].api_response.message_id == mid) {
              convo.sent[s].delivered = true;
          }
        }
      }
    });
  }
});
```

ouadie-lahdioui:
@jsalwen you subscribe to ````message_deliveries```` webhook event ?

ouadie-lahdioui:
I dont know if a have a local problem or facebook doesn't send a delivery confirmation !

[...]

PR #1213, opened 2018-01-15
#1212 check api_response mid or message_id

jsalwen:
fix for issue #1212
check api_response.mid and api_response.message_id for a value to match mid
* `message_id` is currently be returned by facebook
* `mid` might have been returned before; leaving it is case it changes back

### DIFF
--- a/lib/Facebook.js
+++ b/lib/Facebook.js
@@ -10,33 +10,6 @@ function Facebookbot(configuration) {
     // Create a core botkit bot
     var facebook_botkit = Botkit(configuration || {});
 
-    if (facebook_botkit.config.require_delivery) {
-
-        facebook_botkit.on('message_delivered', function(bot, message) {
-
-            // get list of mids in this message
-            for (var m = 0; m < message.delivery.mids.length; m++) {
-                var mid = message.delivery.mids[m];
-
-                // loop through all active conversations this bot is having
-                // and mark messages in conversations as delivered = true
-                bot.findConversation(message, function(convo) {
-                    if (convo) {
-                        for (var s = 0; s < convo.sent.length; s++) {
-                            if (convo.sent[s].sent_timestamp <= message.delivery.watermark ||
-                              (convo.sent[s].api_response &&
-                                  (convo.sent[s].api_response.mid || convo.sent[s].api_response.message_id) == mid)) {
-                                convo.sent[s].delivered = true;
-                            }
-                        }
-                    }
-                });
-            }
-
-        });
-
-    }
-
     // For backwards-compatability, support the receive_via_postback config option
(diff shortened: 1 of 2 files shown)


jsalwen:
@benbrown please review asap, as this is a blocking bug

I'm not sure if facebook changed, but in their documentation, it shows `message_id`
https://developers.facebook.com/docs/messenger-platform/reference/send-api#response

peterswimm:
@ouadie-lahdioui @jonchurch Could we also get your feedback

lib/Facebook.js
```diff
@@ -24,7 +24,8 @@ function Facebookbot(configuration) {
                     if (convo) {
                         for (var s = 0; s < convo.sent.length; s++) {
                             if (convo.sent[s].sent_timestamp <= message.delivery.watermark ||
-                                (convo.sent[s].api_response && convo.sent[s].api_response.mid == mid)) {
+                              (convo.sent[s].api_response &&
+                                  (convo.sent[s].api_response.mid || convo.sent[s].api_response.message_id) == mid)) {
```
ouadie-lahdioui: I don't see where ````api_response.mid```` is used ! maybe it should be removed !

jsalwen: @ouadie-lahdioui I don't see where `api_response.mid` is used either, but it was there.

I'm unclear if this ever worked?!? I can remove the check for `api_response.mid`, but I don't want to do it unless directed, since there might have been a reason for it in the past.

jsalwen:
@ouadie-lahdioui @peterswimm @jonchurch

I can remove the check for `api_response.mid` if  you want. I'm unclear if `require_delivery` ever worked for facebook messenger. If it did, I'm hesitant to remove the `api_response.mid` check.

jsalwen:
hold off on this. I am now testing with the latest version of botkit, and it isn't working.

jsalwen:
@ouadie-lahdioui please review and test again.

the code at the top that I removed wasn't being called in the updated botkit code. I *think* using the `facebook_botkit.middleware.receive.use` is the best way to put this code.

Also, it does require being subscribed to the `message_deliveries` webhook event:
https://developers.facebook.com/docs/messenger-platform/reference/webhook-events/message-deliveries

that event returns:
```js
{
  "sender":{
    "id":"<PSID>"
  },
  "recipient":{
    "id":"<PAGE_ID>"
  },
   "delivery":{
      "mids":[
         "mid.1458668856218:ed81099e15d3f4f233"
      ],
      "watermark":1458668856253,
      "seq":37
   }
}  
```

benbrown:
thank you for the contribution @jsalwen and thanks for your patience while it was reviewed and merged.

I'll be including this in the next release!

PR #1225, opened 2018-01-24
Call FaceBook call back functions if error has occurred

ouadie-lahdioui:
Hello,

While error in getting tags list, Botkit doesn't call cb function with the error. this PR allows ````tags.get_all```` to handle the cb function with ````err```` parameter, it also check if ````cb```` exists before calling it.

Enjoy,


benbrown:

PR #1231, opened 2018-01-25
[Typescript] Rename SocketBot to WebBot

Naktibalda:
Last minute change in webchannel branch renamed SocketBot to WebBot in javascript code,
I implemented the same change in Typescript definition file.


benbrown:

PR #1232, opened 2018-01-29
Add updateResponse callback to replyAndUpdate type

### LINKED ISSUE #1229: Typescript typing for bot.replyAndUpdate
The current type:

https://github.com/howdyai/botkit/blob/6f6ee8c8d88cf5c2e286094b232d477836ec6401/lib/Botkit.d.ts#L257

I think it should be something like:

```ts
replyAndUpdate(src: SlackMessage, resp: string | SlackMessage, cb: (err: Error, res: string, updateResponse: (newResp: string | SlackMessage, cb?: (err: Error) => void) => void): void;
```

This contains the extra `updateResponse` callback found in the docs: https://github.com/howdyai/botkit/blob/master/docs/readme-slack.md#botreplyandupdate

ThomWright:
Fixes #1229

### DIFF
--- a/lib/Botkit.d.ts
+++ b/lib/Botkit.d.ts
@@ -244,6 +244,9 @@ declare namespace botkit {
     title_link?: string;
     ts?: string;
   }
+  interface SlackUpdateMessageCallback {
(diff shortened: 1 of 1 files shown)


benbrown:

PR #1233, opened 2018-01-29
should use value not payload as keyname

auberryberry:
@peterswimm @benbrown fix for openUrl action of the hero cards, doesn't currently work but will when you use value over payload for the keyname.


benbrown:

PR #1252, opened 2018-02-14
Smooch.io

hwhelchel:
This is a basic implementation of Smooch support for Botkit. I've tested it to ensure quick replies work in the Smooch Web Messenger. I haven't tested other message types or channels.

screen shot 2018-02-13 at 9 25 14 pm


ryansydnor:
The `legacy` branch is no longer accepting new platform integrations. If you'd like to port this to `v4`, please create a `botbuilder-adapter-smooch` (similar to the other `botbuilder-adapter-*` examples, found here: https://github.com/howdyai/botkit/tree/master/packages)

PR #1264, opened 2018-02-24
Added 'picture' to Facebook User Profile endpoint.

se:
I have added profile picture information to Facebook User Profile.


julianusti:
@se I think you have some formatting issues with your PR. On top of that, there is a function called [getMessageUser()](https://github.com/howdyai/botkit/blob/master/lib/Facebook.js#L231) which I think should be extended with `picture` prop as well, but I'm not sure about the last one, so maybe @benbrown can say more here.

benbrown:
Merged! Thank you.

PR #1265, opened 2018-02-24
Extend Facebook's user profile object with locale.

julianusti:
There is a missing field, namely`locale` which is a part of the identity response object of `bot.getMessageUser`

### DIFF
--- a/lib/Facebook.js
+++ b/lib/Facebook.js
@@ -241,6 +241,7 @@ function Facebookbot(configuration) {
                         full_name: identity.first_name + ' ' + identity.last_name,
                         email: identity.email || null,
                         gender: identity.gender,
+                        locale: identity.locale,
                         timezone_offset: identity.timezone,
                     };
 


julianusti:
@benbrown please review. I'm also kind of confused with the name of this function `getMessageUser` since I use it to **getUserProfile**.

julianusti:
@benbrown @ouadie-lahdioui @peterswimm, Sorry guys for intolerance and tag all of you 😃 this is a critical issue for us, to go live upcoming Monday, and I would love to know if this fix makes sense or not or maybe there is another way to identify user's locale? Thanks in advance.

peterswimm:
How is making this public blocking you going live? You can always just publish to your own repro and go live against that.

I will ping our team and see if they can review, but we were not planning a release in your timeframe.

julianusti:
@peterswimm that's what we did. Thx for reply. Looking forward to PR review.

peterswimm:
@julianusti The latest version has been fixed, let us know if you run into any issues.

PR #1280, opened 2018-03-07
Send broadcast function: added missing args(notification and tag)

### LINKED ISSUE #1271: Send broadcast message missing notification_type and tag
The API in facebook docs look like this:
screen shot 2018-03-04 at 2 20 16 pm

Where the function in facebook.js only include the creative_message_id

screen shot 2018-03-04 at 2 19 08 pm

I can work on it if needed but we first have to agree on how the pass the missing two parameters.
I think we should convert the first param to an object.

OmranAbazid:
To fix #1271

### DIFF
--- a/lib/Facebook.js
+++ b/lib/Facebook.js
@@ -1195,11 +1195,7 @@ function Facebookbot(configuration) {
                     'message_creative_id': message_creative
                 };
             } else {
-                body = {
-                    'message_creative_id': message_creative.message_creative_id,
-                    'notification_type': message_creative.notification_type,
-                    'tag': message_creative.tag,
-                };
+                body = message_creative;
             }
 
             if (custom_label_id) {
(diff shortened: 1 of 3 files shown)


lib/Facebook.js
```diff
@@ -1182,16 +1182,25 @@ function Facebookbot(configuration) {
...
 
-            var body = {
-                'message_creative_id': message_creative_id
-            };
+            if (typeof message_creative === 'string') {
+                body = {
+                    'message_creative_id': message_creative
+                };
+            } else {
+                body = {
```
ouadie-lahdioui: It will be smart to clone all message_creative attributes into the body instead of copying one by one :

````
} else {
    body = message_creative
}
````

OmranAbazid:
Good point, updated the code.

ouadie-lahdioui:
LGTM 👍 Thank you :wink:

PR #1283, opened 2018-03-12
Fix broken Dashbot links in middleware docs

urbanslug:
Fix broken links to dashbot botkit middleware (Slack and Facebook) that leads to a 404.


peterswimm:
Thanks!!

PR #1301, opened 2018-03-22
Fix #1279

### LINKED ISSUE #1279: quick reply with content type email or phone returns error
hi im trying to use the quick reply for email or phone in the way below
    convo.addQuestion({
                        "text": "Here is a quick reply!",
                        quick_replies: [
                            {
                                content_type: 'user_email',
                            },
                            {
                                content_type: 'user_phone_number',
                            },
                        ],
                    }
and getting this error
nfo: An error occurred while sending a message:  { message: '(#100) The parameter message[quick_replies][0][content_type] is required',
  type: 'OAuthException',
  code: 100,
  fbtrace_id: 'DqOcSLT3A8I' }

if i use content types like text and location they are working without any issue

ouadie-lahdioui:
Hello,

In this PR i replaced the internat core quick replies format mechanism by quick replies from message object.

This way will prevent us adding new supported quick replies each time fb adds ones.

FYI FB truncates the quick reply title by default, i also removed ````titleLimit````

WDYT ?

### DIFF
--- a/lib/Facebook.js
+++ b/lib/Facebook.js
@@ -55,37 +55,8 @@ function Facebookbot(configuration) {
                 platform_message.message.sticker_id = message.sticker_id;
             }
 
-            if (message.quick_replies) {
+            if (message.quick_replies) platform_message.message.quick_replies = message.quick_replies;
 
-                // sanitize the length of the title to maximum of 20 chars
-                var titleLimit = function(title) {
-                    if (title.length > 20) {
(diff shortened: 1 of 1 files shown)


benbrown:
I think this mostly makes sense.

To avoid breaking anybody, could you loop through the parameters as provided in message and ensure they have content_type set, and if not set it to text?  This lets people exclude that field in the simplest cases.

ouadie-lahdioui:
Done 👍

PR #1312, opened 2018-04-04
FB require_delivery: Remove use of message.delivery.mids because not guaranteed

### LINKED ISSUE #1311: `require_delivery` sometimes fails on Facebook
We [currently](https://github.com/howdyai/botkit/blob/9a876e6dc2686117b3cdc7fb1e6f6ed74e0d9dd0/lib/Facebook.js#L515) assume that `message.delivery.mids` always exists. This is not always the case, as specified in [Facebook's docs](https://developers.facebook.com/docs/messenger-platform/reference/webhook-events/message-deliveries):

**Property** | **Type** | **Description**
--------------|----------------|-----------------
   `mids`    |     Array     | Array containing message IDs of messages that were delivered. *Field may not be present.*

(italics mine)

In terms of a fix, what do you think of iterating through all messages and marking as "delivered" the ones after the given "watermark" date?

cchan:
Fixes #1311

### DIFF
--- a/lib/Facebook.js
+++ b/lib/Facebook.js
@@ -511,24 +511,19 @@ function Facebookbot(configuration) {
     facebook_botkit.middleware.receive.use(function handleDelivery(bot, message, next) {
 
         if (message.type === 'message_delivered' && facebook_botkit.config.require_delivery) {
-            // get list of mids in this message
-            for (var m = 0; m < message.delivery.mids.length; m++) {
-                var mid = message.delivery.mids[m];
-
-                // loop through all active conversations this bot is having
-                // and mark messages in conversations as delivered = true
-                // note: we don't pass the real event in here because message_delivered events are excluded from conversations and won't ever match!
-                bot.findConversation({user: message.user}, function(convo) {
-                    if (convo) {
-                        for (var s = 0; s < convo.sent.length; s++) {
-                            if (convo.sent[s].sent_timestamp <= message.delivery.watermark ||
-                          (convo.sent[s].api_response && convo.sent[s].api_response.message_id == mid)) {
-                                convo.sent[s].delivered = true;
-                            }
+            // loop through all active conversations this bot is having
+            // and mark messages in conversations as delivered = true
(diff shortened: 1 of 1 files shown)


benbrown:
@cchan can you help me understand how to test this?

cchan:
@benbrown That's an interesting question; I'm not sure how testing works for botkit. Do you have mocks set up for Facebook?

cchan:
(oops)

cchan:
@benbrown Hm, could we create a sinon test similar to [this existing one](https://github.com/howdyai/botkit/blob/master/tests/Facebook.js#L165)? I'm not familiar at all with how botkit works internally, but it seems like there's a polling loop that checks if the previous message has been marked as delivered before sending the next one.

benbrown:
@cchan my original comment probably confused the issue. what I need to know is what settings to enable  and what type of events I need to trigger in order to see this fix in action.

cchan:
@benbrown Ah, that's what you mean. You just need to set `require_delivery: true` and send a bunch of messages in a row from Botkit - that's pretty much it. Facebook will respond with delivery events, but once in a while the response will not contain the field `mids`, which crashes the old code.

I haven't been able to figure out what conditions cause Facebook to not provide `mids` in the response, but after you try it for enough times / long enough it'll just happen, seemingly at random. I think it's probably a load reduction feature on Facebook's end - the "delivered" feature is probably fundamentally based on the `watermark` timestamp, and when they don't have the capacity to generate the corresponding list of message IDs, they simply don't.

(The reason I found this bug is that I have a bot on my server (`clive-io/statbot`) that monitors logs et cetera, and it was crashing at random about once a week.)

RJiryes:
Hey is this going to be merged anytime soon? Great catch @cchan ;-)

benbrown:
@cchan @RJiryes merged!

cchan:
Awesome! :D

RJiryes:
@benbrown Thanks!

PR #1318, opened 2018-04-11
Updating typings for interactive messages

### LINKED ISSUE #1317: Missing Typescript Types for Slack Interactive Messages
Botkit Version: 0.6.13

Building Typescript Slack Bot using Botkit and ran into issues regarding type checking:

SlackAttachment interface missing the following:
- actions
- attachment_type
- callback_id

Message interface missing the following:
- replace_original
- response_type

Type check errors all follow a similar error message:

` Object literal may only specify known properties, and 'callback_id' does not exist in type 'SlackAttachment'.`

Was able to workaround the issue with either of the two workarounds below:

1) Added missing types locally to https://github.com/howdyai/botkit/blob/master/lib/Botkit.d.ts
2) Changed the ts file to a js file and updated my compiler to allowJs.

jaalger:
- Updating SlackAttachment type for interactive messages
- Updating Message type for interactive message handling

closes #1317

### DIFF
--- a/lib/Botkit.d.ts
+++ b/lib/Botkit.d.ts
@@ -221,13 +221,18 @@ declare namespace botkit {
     action?: string;
     channel?: string;
     match?: RegExpMatchArray;
+    replace_original?: boolean;
+    response_type?: string;
     text?: string;
     user?: string;
   }
   interface SlackAttachment {
+    attachment_type?: string;
+    actions?: any[];
     author_icon?: string;
     author_link?: string;
     author_name?: string;
+    callback_id?: string;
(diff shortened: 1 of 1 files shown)


lib/Botkit.d.ts
```diff
@@ -221,13 +221,18 @@ declare namespace botkit {
     action?: string;
     channel?: string;
     match?: RegExpMatchArray;
+    replace_original?: boolean;
+    response_type?: string;
```
Naktibalda: These properties are Slack specific, please move them to SlackMessage interface.

Issue #1244, opened 2018-02-07
Facebook quick_replies payload is not being received in postback 


MagSag:
My quick reply button has the following structure : 
```
{
            "text": t(language,"Would you like to get it ?"),
            "quick_replies": [
                {
                    "content_type": "text",
                    "title": "YES",
                    "payload": "GETYES",
                },
                {
                    "content_type": "text",
                    "title": "NO",
                    "payload": "GETNO",
                },
            ]
        }
```

And my postback handler : 
		
```
		controller.on('facebook_postback', function (bot, message) {
		if (message.payload==='GETYES'){
		bot.reply(message,'Thank you!!!');
		}else if (message.payload==='GETNO') {
		  bot.reply(message,'Why no?');
		}
		
		});

The user is receiving nothing after clicking YES or NO.
```

nicolasjust:
Hi
I have the same error, how can I fix?

ouadie-lahdioui:
Hello,

Try to catch them as a ````message_received```` instead of ````facebook_postback````

NatpakalJAM:
I have same problem, Can someone help me.

My quick reply button has the following structure :
```
controller.hears('answer_finished', 'facebook_postback', async function (bot, message) {

        let lang = await CheckLang.checkLang(bot.api.token);
        await bot.reply(message, {
            text: 'Yes or No',
            quick_replies: [{
                    'content_type': 'text',
                    'title': 'Yes',
                    'payload': 'GETYES'
                },
                {
                    'content_type': 'text',
                    'title': 'No',
                    'payload': 'GETNO'
                }
            ]
        });
    });
```

And my postback handler :

```
controller.on('message_received', async(bot, message) => {
        await bot.reply(message, `Yes or No`);
        await bot.reply(message, `Echo: ${ message.payload }`);
    });
```

I try to use all event in [Doc V4+](https://botkit.ai/docs/v4/reference/facebook.html#facebookeventtypemiddleware) but it still not working

PR #1342, opened 2018-05-07
Add native facebook "quick_reply" support as "facebook_quick_reply"

brh55:
I've noticed that the quick_reply is perceived as a "message_received" event, where as the "facebook_postback" is a native event type. This is confusing for initial users unless time is spent reading through issues or the source code (#1224 + including myself).

This pull request adds "quick_reply" as a native facebook event. I've also assigned the `quick_reply.payload` to `message.text` as this would make it easier for creating proper listeners when combined with RegEx versus the quick_reply button text.

Example:
```js
// Can listen to both types of payloads with similar payloads
controller.hears("foo:.*", ['facebook_quick_reply', 'facebook_postback'], (bot, message) => {
    bot.reply(message, "You say foo, I say bar");
});

### DIFF
--- a/docs/readme-facebook.md
+++ b/docs/readme-facebook.md
@@ -89,10 +89,10 @@ Normal messages will be sent to your bot using the `message_received` event.  In
 | facebook_quick_reply | A user clicked a "quick reply" and triggered a quick reply message.
 | message_delivered | A confirmation from Facebook that a message has been received.
 | message_echo | If enabled in Facebook, an "echo" of any message sent by the bot.
-| message_read | A confirmation from Facebook that a message has been read
-| facebook_account_linking | A user has started the account linking
-| facebook_optin | A user has clicked the [Send-to-Messenger plugin](https://developers.facebook.com/docs/messenger-platform/implementation#send_to_messenger_plugin).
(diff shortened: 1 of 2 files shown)


benbrown:
This change would break a bunch of existing bots that expect quick replies to act like normal messages.

I would suggest implementing this as a plugin middleware that is separate from Botkit Core.

Can I ask, why do you want it to be a special event?

brh55:
@benbrown 
That's a good question. I suppose I've wrongfully viewed quick replies as a in-between offering to text and buttons (postbacks) due to the custom payload similar to a postback. Thus the complexity arises when a quick reply label (which is considered `message.text`) may have different utterances with identical payloads, or my payloads have identical labels with slightly different payloads (`suggest:<user>`, `suggest:<business>`, etc). 

However, it seems this is a common confusion among users (#1244 , #1013, #992), so perhaps at a minimum an example general middleware in the docs would greatly benefit users as well as prevent breaking changes. 

It seems there was a similar PR to this #992 (my apologize for duplication) suggesting a backward compatible solution triggering both events, but it would probably make sense to treat specific quick_reply event using the `quick_replies.payload` as the `message.text`, and maintain the default ` message_received` behavior as the `quick_reply.text` as the `message.text` to prevent duplication in triggered events and maximum compatibility.

benbrown:
We are currently not going to change the way quick replies are handled.

PR #1359, opened 2018-05-25
Add missing methods to d.ts

ypresto:
I want to use `say()` without casting bot to `any`. 😃
Also fixes `BotFrameworkController.createWebhookEndpoints()` accepts `TwilioSMSBot` by mistake.

### DIFF
--- a/lib/Botkit.d.ts
+++ b/lib/Botkit.d.ts
@@ -2,13 +2,15 @@ import * as express from 'express';
 import * as http from "http"
 
 declare namespace botkit {
-  function botframeworkbot(configuration: BotFrameworkConfiguration): BotFrameworkController;
-  function consolebot(configuration: ConsoleConfiguration): ConsoleController;
-  function facebookbot(configuration: FacebookConfiguration): FacebookController;
   function slackbot(configuration: SlackConfiguration): SlackController;
   function sparkbot(configuration: CiscoSparkConfiguration): CiscoSparkController;
+  function facebookbot(configuration: FacebookConfiguration): FacebookController;
   function twilioipmbot(configuration: TwilioIPMConfiguration): TwilioIPMController;
   function twiliosmsbot(configuration: TwilioSMSConfiguration): TwilioSMSController;
+  function botframeworkbot(configuration: BotFrameworkConfiguration): BotFrameworkController;
+  function teamsbot(configuration: TeamsConfiguration): TeamsController;
+  function consolebot(configuration: ConsoleConfiguration): ConsoleController;
+  function jabberbot(configuration: JabberConfiguration): JabberController;
   function socketbot(configuration: WebConfiguration): WebController;
   function anywhere(configuration: WebConfiguration): WebController;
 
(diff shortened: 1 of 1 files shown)


lib/Botkit.d.ts
```diff
@@ -465,6 +489,19 @@ declare namespace botkit {
   interface Team {
     id: string;
   }
+  interface TeamsBot extends Bot<TeamsSpawnConfiguration, TeamsMessage> {
+  }
+  interface TeamsConfiguration extends Configuration {
+    clientId?: string;
+    clientSecret?: string;
```
Naktibalda: Teams bot has more configuration options:
https://github.com/howdyai/botkit/blob/master/lib/Teams.js#L358-L368
https://github.com/howdyai/botkit/blob/master/lib/Teams.js#L209

lib/Botkit.d.ts
```diff
@@ -217,6 +225,22 @@ declare namespace botkit {
     name: string;
     emails: string[];
   }
+  interface JabberBot extends Bot<JabberSpawnConfiguration, JabberMessage> {
+  }
+  interface JabberConfiguration extends Configuration {
+  }
+  interface JabberController extends Controller<JabberSpawnConfiguration, JabberMessage, JabberBot> {
+  }
+  interface JabberMessage extends Message {
```
Naktibalda: I see that JabberMessage has custom fields stanza and group, please add them to this interface.
https://github.com/howdyai/botkit/blob/master/lib/JabberBot.js#L192-L195

from_jid field could be useful too.

lib/Botkit.d.ts
```diff
@@ -465,6 +489,19 @@ declare namespace botkit {
   interface Team {
     id: string;
   }
+  interface TeamsBot extends Bot<TeamsSpawnConfiguration, TeamsMessage> {
```
Naktibalda: TeamsBot has many custom methods:
https://github.com/howdyai/botkit/blob/master/lib/Teams.js#L43-L325

lib/Botkit.d.ts
```diff
@@ -465,6 +489,19 @@ declare namespace botkit {
...
+  }
+  interface TeamsConfiguration extends Configuration {
+    clientId?: string;
+    clientSecret?: string;
+  }
+  interface TeamsController extends Controller<TeamsSpawnConfiguration, TeamsMessage, TeamsBot> {
+    createWebhookEndpoints(): this;
+  }
+  interface TeamsMessage extends Message {
```
Naktibalda: TeamsMessage has many custom fields.
https://github.com/howdyai/botkit/blob/master/lib/Teams.js#L118-L120
https://github.com/howdyai/botkit/blob/master/lib/Teams.js#L581-L588

Naktibalda:
This pull request is a good addition on its own and can be merged, but somebody will have to add missing parts of Jabber and Teams bots later.

ypresto:
Thanks for reviewing! I'll work on this this week end.

ypresto:
I worked for hours on TeamsBot, and found some of existing code / doc has possible bugs or outdated things.

For example, in helper methods of attachment object (returned from `createHero()` and `createThumbnail()`), the `tap(type: string, title: string, payload: string)` method should be `tap(type: string, value: string)` according to [MS official reference](https://docs.microsoft.com/en-us/microsoftteams/platform/concepts/cards/cards-reference).

Also, [botkit's doc](https://github.com/howdyai/botkit/blob/4651e1491f2bfea67263b6ad97c0c06d378eb365/docs/readme-teams.md#100) says `The serviceUrl can be extracted from the incoming message payload at message.serviceUrl`, but it would be `message.raw_message.serviceUrl` according to [code]( https://github.com/howdyai/botkit/blob/dc0e780d3a50ffbfe89bc8f3908d1f8869d61466/lib/Teams.js#L168).

It's outside of scope of this PR so I didn't fixed them.

Anyway it costs too much for me, I gave up writing Jabber's interface...

PR #1362, opened 2018-05-30
Add new users.conversations web api endpoint

pthieu:
Back with another one, 2 years later.

New method came out in April 2018.


benbrown:
THANKS!

PR #1370, opened 2018-06-02
Added select dialog element with data source

drwho725:
Allow select dialog options to be populated from data sources: https://api.slack.com/dialogs#populate_a_select_menu_dynamically


drwho725:
Closing the PR, since addSelect supports `data_source` just have to pass it via options parameter 🤦‍♂️ .

Issue #1386, opened 2018-06-14
Feature Request: Functionality to manipulate the typingDelay time globally


schmitzl:
I want to manipulate the time for the typingDelay for my SocketBot. However, I would have to set the typingDelay for every message individually. This is far too much work for something, that could as well be solved with a configuration variable. I do not want to change the original formula to calculate the typingDelay as it works nicely and takes the message length into account. What I want instead is an additional factor in that calculation, that I can set for bot. This typingDelayFactor should be 1 by default and with this use the original calculation for the typingDelay. However, as a user, I can also increase the typing speed of the bot by setting it to something lower than 1 or I can turn my bot into a really slow typer by setting it to e.g. 5. This gives a lot of flexibility while still taking the message length into account. 
I will implement the changes for a typingDelayFactor and create a pull request.

PR #1387, opened 2018-06-14
#1386: Additional global parameter that can be set to change the typi…

schmitzl:
Ticketnummer: #1386

## controller.setTypingDelayFactor();

Adjust the speed of the typing delay by setting a factor. The typing delay determines the time that the bot takes, to reply to a message. By default, the factor is set to 1, using a predefined calculation for the typing delay, that takes the message length into account. In order to increase or decrease the typing delay, the factor can be changed. For example, the default typing delay can be doubled by setting the factor to 2.

**For example:**

```
// double the time of the typing delay
controller.setTypingDelayFactor(2);
```

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -31,7 +31,6 @@ function Botkit(configuration) {
             teams: {},
         },
         tickDelay: 1500,
-        typingDelayFactor: 1,
     };
 
     /* ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
@@ -1437,11 +1436,6 @@ function Botkit(configuration) {
         if (botkit.tickInterval) {
             clearInterval(botkit.tickInterval);
         }
-    };
-
-    // change the speed of typing a reply in a conversation
-    botkit.setTypingDelayFactor = function(delayFactor) {
-        botkit.typingDelayFactor = delayFactor;
     };
 
     botkit.startTask = function(bot, message, cb) {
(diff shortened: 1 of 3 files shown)


benbrown:
@schmitzl I think that since this only affects the web adapter, it should be part of that file, not in Core.

I try to avoid putting anything in core that doesn't apply to all platforms.

Can you define the function and variable in Web.js instead?

Also, please add a note about this to the docs page for the web connector that is contained in [this other repo](/howdyai/botkit-docs)

lib/CoreBot.js
```diff
@@ -1438,6 +1439,11 @@ function Botkit(configuration) {
         }
     };
 
+    // change the speed of typing a reply in a conversation
+    botkit.setTypingDelayFactor = function(delayFactor) {
```
benbrown: define this in lib/Web.js instead

lib/CoreBot.js
```diff
@@ -31,6 +31,7 @@ function Botkit(configuration) {
             teams: {},
         },
         tickDelay: 1500,
+        typingDelayFactor: 1,
```
benbrown: add this to the options in web.js ony

schmitzl:
@benbrown Thank you for the feedback. I implemented the changes and created a pull request for the docs (https://github.com/howdyai/botkit-docs/pull/9).

benbrown:
Great, I will re-review! Thanks contributing!

benbrown:
oops @schmitzl it doesn't look like your new changes have been pushed to this PR yet

schmitzl:
@benbrown Sorry, wrong branch! I hope you can view the changes now.

PR #1406, opened 2018-07-02
Request Signing with Slack Signing Secret

shishirsharma:
This is first cut. 

Biggest caveat in the current approach is control of webserver. HMAC signature needs raw request body. So far the only way to do with express is via a middleware. That needs to be added before usual `handleWebhookPayload`.

PS: You will have to add the following snippet just after you create `webserver` in your code.
```js
  var webserver = express();
  webserver.use(function(req, res, next) {
    req.rawBody = '';
    
    req.on('data', function(chunk) {
      req.rawBody += chunk;
    });
    
    next();
  });
```

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -1488,6 +1488,15 @@ function Botkit(configuration) {
         botkit.config.port = port;
 
         botkit.webserver = express();
+        botkit.webserver.use(function(req, res, next) {
+            req.rawBody = '';
+
+            req.on('data', function(chunk) {
+                req.rawBody += chunk;
+            });
+
+            next();
+        });
         botkit.webserver.use(bodyParser.json());
         botkit.webserver.use(bodyParser.urlencoded({
             extended: true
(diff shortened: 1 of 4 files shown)


chrilian:
Hi there-- I tried this in my code and the req.rawBody comes out as empty? Is this just me? nothing gets passed into it in the stream basically...

shishirsharma:
Yes, You will have to add the following snippet just after you create `webserver` in your code. 
```js
  var webserver = express();
  webserver.use(function(req, res, next) {
    req.rawBody = '';
    
    req.on('data', function(chunk) {
      req.rawBody += chunk;
    });
    
    next();
  });

```

chrilian:
my bad-- I put it in the wrong place

lib/CoreBot.js
```diff
@@ -1488,6 +1488,15 @@ function Botkit(configuration) {
         botkit.config.port = port;
 
         botkit.webserver = express();
+        botkit.webserver.use(function(req, res, next) {
```
benbrown: Can you make this same change to the slack starter kit?  github.com/howdyai/botkit-starter-slack

shishirsharma: Sure

PR #1421, opened 2018-07-20
fix: add return types for TeamsAPI method type definitions

### LINKED ISSUE #1420: Typescript with noImplicitAny fails with newly added TeamsAPI definition.
**Version:** `0.6.15+`
**OS:** `OSX High Sierra 10.13.6` (but this will effect all OS)

**What**
Upgrading to `0.6.15` introduces a new type definition for `TeamsAPI` - these do not have return types defined, so having `noImplicitAny: true` in the `tsconfig.json` causes the typescript compiler to complain. This was [introduced here](https://github.com/howdyai/botkit/commit/3b332da0808a8cbe77628b4c13e8ddf575679171#diff-b3264e84da512f486da2e2a6efbf2e78R492) and they should all have a return type of `void` as they all execute a callback.

screen shot 2018-07-20 at 10 30 19 am

alexframe:
This pull request adds return types to the type definitions for the `TeamsAPI` interface.

Resolves #1420.

### DIFF
--- a/lib/Botkit.d.ts
+++ b/lib/Botkit.d.ts
@@ -496,12 +496,12 @@ declare namespace botkit {
     id: string;
   }
   interface TeamsAPI {
-    getUserById(conversationId: string, userId: string, cb: (err: Error, user_profile: any) => void);
-    getUserByUpn(conversationId: string, upn: string, cb: (err: Error, user_profile: any) => void);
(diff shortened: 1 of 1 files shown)


benbrown:
thanks!

alexframe:
@benbrown would it be possible to publish this to npm when you have some time? 🙏

PR #1425, opened 2018-07-23
Add receive_error event if middleware throws error

Nop0x:
Previously, if an error occurred in the receive middleware, the message would just get swallowed and not processed. The error is just logged to console. Now an actual event is fired which can be caught in a skill.

- CoreBot.js: firing receive_error event with the error message of the middleware
- Error can be caught by using `controller.on('receive_error', function(err, bot, message){[HANDLE THE ERROR...]});`


benbrown:
This is merged into my next release branch!

PR #1437, opened 2018-08-01
replyWithTyping in botkit studio

watzing:
Allow messages in Botkit Studio to display the typing indicator if a typing custom field is set to true


benbrown:
This might be better implemented as a plugin middleware rather than a change to core.

For example:
```
controller.middleware.send.use(function(bot, message, next) {
if (message.typing) {
bot.send({typing:true});
setTimeout(function() {
next() },2000);
} else {
next();
});
```

Issue #1428, opened 2018-07-25
Slack initiate conversation from external event


Oliverr:
Our SlackBot receives a notification from an external source which is then driven into a bot.say() event to the Slack channel.
When someone responds to the output of bot.say() we want it to drive a further action back to the external source (via a POST using some of the message in the response + original message).
Sample flow:

```
BOT: XYZ has just occurred on ABC
USER: Tell me more
BOT: Here are the latest details for the event
```

Is bot.say() the correct method to initiate this workflow?
The conversation will be open to the entire channel, not an individual user.

Oliverr:
@benbrown confirmed that message_replied is returning a non-normalised structure which is not currently supported.
Work around is:

```
controller.middleware.categorize.use(function(bot, message, next) {
if (message.type=='message_replied') {
 message.text = message.message.text;
 // and copy user and other fields 
} 
next() });
```
Please mark as issue/bug

peterswimm:
Will do, will you be submitting a PR?

Oliverr:
I just have: #1441

ryansydnor:
Given the `legacy` branch is in maintenance mode and the associated PR is stale - I'm going to close this issue.

PR #1441, opened 2018-08-03
Add handler for message_replied

Oliverr:
message_replied is not currently supported as it has a non-normalised structure

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -394,6 +394,10 @@ function Slackbot(configuration) {
 
 
             }
+        } else if (message.type == 'message_replied') {
+            message.text = message.message.text;
+            message.user = message.message.user;
+            message.replies = message.message.replies;
         }
 
         // move on to the next stage of the pipeline
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -334,6 +334,7 @@ function Slackbot(configuration) {
 
             if (message.text) {
                 message.text = message.text.trim();
+                next();
             }
 
             // set up a couple of special cases based on subtype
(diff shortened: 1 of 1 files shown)


Oliverr:
Resolution for #1428

benbrown:
After some research on Slack's site, I think this is not quite right.

The text included in the message.message.text field is actually the PARENT MESSAGE, not the reply.

I think we need to actually call `conversations.replies` and get the reply text and then use that for the hears comparisons.

Oliverr:
Work in progress. Not happy yet, had to relate the next() calls to allow for API query completion before moving on

```
}  else if (message.subtype == 'message_replied') {
                message.type = 'message_replied';
                message.text = message.message.text;
                message.user = message.message.user;
                var options = { method: 'POST',
                    url: 'https://slack.com/api/conversations.replies',
                    formData: 
                        { token: slack_botkit.config.clientSecret,
                         channel: message.channel,
                         ts: message.message.replies[(message.message.replies).length - 1].ts }
                    };
                request(options, function (error, response, body) {
                    message._requestreply = JSON.parse(body).messages;
                    next();
                });
            }
```

Oliverr:
@peterswimm @benbrown proposed changes committed to PR. Interested in your thoughts on the need for next() to be relocated...

Oliverr:
@peterswimm @benbrown any thoughts on this?

lib/SlackBot.js
```diff
@@ -344,15 +345,35 @@ function Slackbot(configuration) {
...
                 } else {
                     message.type = 'user_group_join';
                 }
-            } else if (message.subtype) {
+                next();
+            }  else if (message.subtype == 'message_replied') {
+                message.type = 'message_replied';
+                message.text = message.message.text;
+                message.user = message.message.user;
+                var options = { method: 'POST',
```
benbrown: this should use bot.api.conversations.replies instead of making a direct call to request.

other than that, I think this looks good to go.

benbrown: @Oliverr Want to take a last pass at this so we can merge this in?

Issue #1450, opened 2018-08-27
Socketbot crashes with non JSON incoming message


b-a-byte:
Hi there,

I'm facing an issue with socketbot and non JSON messages.

I'm creating my bot from [botkit-starter-web](https://github.com/howdyai/botkit-starter-web) and NodeJS app appears to crash when I'm sending non JSON message.

This is annoying in production environment :)

Thank you for Botkit, nice tool

(PS: Pull request incoming)

### Reproduction

0. Run `npm start` from [botkit-starter-web](https://github.com/howdyai/botkit-starter-web) template
0. Send a string to `ws://localhost:3000`

    ```javacript
    const websocket = new WebSocket('ws://localhost:3000');
    websocket.addEventListener('open', () => websocket.send('message'));
    ```
0. Boom!

### Logs
```bash
undefined:1
message
^

SyntaxError: Unexpected token m in JSON at position 0
    at JSON.parse (<anonymous>)
    at WebSocket.incoming (/home/labs/botkit-starter-web/node_modules/botkit/lib/Web.js:35:36)
    at emitOne (events.js:116:13)
    at WebSocket.emit (events.js:211:7)
    at Receiver.receiverOnMessage (/home/labs/botkit-starter-web/node_modules/botkit/node_modules/ws/lib/websocket.js:720:20)
    at emitOne (events.js:116:13)
    at Receiver.emit (events.js:211:7)
    at Receiver.dataMessage (/home/labs/botkit-starter-web/node_modules/botkit/node_modules/ws/lib/receiver.js:414:14)
    at Receiver.getData (/home/labs/botkit-starter-web/node_modules/botkit/node_modules/ws/lib/receiver.js:346:17)
    at Receiver.startLoop (/home/labs/botkit-starter-web/node_modules/botkit/node_modules/ws/lib/receiver.js:133:22)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! botkit-starter-web@0.0.1 start: `node bot.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the botkit-starter-web@0.0.1 start script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

```

Botkit version: 0.6.16
Operating system: Fedora (28)

simonfranzen:
@gobah Are you working on this issue?

PR #1451, opened 2018-08-27
Prevent app to crash from incoming message

gobah:
Hi,

This is a pull request for this related issue https://github.com/howdyai/botkit/issues/1450

Thank you


benbrown:
Fixed by #1474

Issue #1447, opened 2018-08-20
TypeError when Slack slash command received by bot


sprak3000:
- botkit version `v0.6.16`
- Ubuntu 16.04.5 LTS
- Bot connected to Slack

Had not updated my version of botkit for some time and recently upgraded to v0.6.16. Had a working slash command prior to the upgrade. After the upgrade, sending the slash command results in this error output in the server log:

```
TypeError: Cannot read property 'user_id' of undefined
    at /home/ubuntu/shodan/node_modules/botkit/lib/SlackBot.js:258:34
    at /home/ubuntu/shodan/node_modules/botkit/lib/SlackBot.js:149:17
    at Store._get (/home/ubuntu/shodan/node_modules/botkit/node_modules/jfs/Store.js:198:39)
    at Store.get (/home/ubuntu/shodan/node_modules/botkit/node_modules/jfs/Store.js:351:12)
    at Object.get (/home/ubuntu/shodan/node_modules/botkit/lib/storage/simple_storage.js:54:26)
    at Object.Slackbot.slack_botkit.findTeamById (/home/ubuntu/shodan/node_modules/botkit/lib/SlackBot.js:543:36)
    at Object.Slackbot.slack_botkit.findAppropriateTeam (/home/ubuntu/shodan/node_modules/botkit/lib/SlackBot.js:147:22)
    at Object.Slackbot.slack_botkit.handleWebhookPayload (/home/ubuntu/shodan/node_modules/botkit/lib/SlackBot.js:226:22)
    at /home/ubuntu/shodan/node_modules/botkit/lib/SlackBot.js:136:26
    at Layer.handle [as handle_request] (/home/ubuntu/shodan/node_modules/express/lib/router/layer.js:95:5)
    at next (/home/ubuntu/shodan/node_modules/express/lib/router/route.js:137:13)
    at Route.dispatch (/home/ubuntu/shodan/node_modules/express/lib/router/route.js:112:3)
    at Layer.handle [as handle_request] (/home/ubuntu/shodan/node_modules/express/lib/router/layer.js:95:5)
    at /home/ubuntu/shodan/node_modules/express/lib/router/index.js:281:22
    at Function.process_params (/home/ubuntu/shodan/node_modules/express/lib/router/index.js:335:12)
    at next (/home/ubuntu/shodan/node_modules/express/lib/router/index.js:275:10)
    at serveStatic (/home/ubuntu/shodan/node_modules/express/node_modules/serve-static/index.js:75:16)
[...]

PR #1453, opened 2018-08-30
Address TypeError seen in Issue 1447

sprak3000:
Address Issue #1447 - TypeError when Slack slash command received by bot:
* Move the former `bot.identity` assignment into the `event_callback` payload case
* Add an `else` case to catch non-`event_callback` payloads and set the `bot.identity` appropriately

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -246,20 +246,21 @@ function Slackbot(configuration) {
                 // Recent changes in Slack will break other integrations as they no longer
                 // require a bot and therefore Slack won't send the bot information.
                 if (payload.type === 'event_callback') {
-
                     if (!team.bot) {
                         slack_botkit.log.error('No bot identity found.');
                         return res.status(500).send();
                     }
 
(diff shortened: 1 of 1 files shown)


lib/SlackBot.js
```diff
@@ -246,20 +246,21 @@ function Slackbot(configuration) {
...
                         slack_botkit.log.error('No bot identity found.');
                         return res.status(500).send();
                     }
 
+                    bot.identity = {
+                        id: team.bot.user_id,
+                        name: team.bot.name
+                    };
+                } else {
+                    bot.identity = {
```
sprak3000: Generated the original issue using a bot attached to a Slack app. Used same bot to create this fix. When sending a slash command to my bot, the `team` object was in the form `{ id: 'XXXXXXXXX', name: 'foo', domain: 'foo' }`, and there was no `payload.type` defined.

Based on my understanding of legacy versus Slack connected bots, I believe this approach should handle both types. Using this patch with my bot restores slash command functionality.

PR #1476, opened 2018-10-08
add slack channel type

casz:


peterswimm:
Can you explain more your thinking behind this pull request?

casz:
A slack message from slash command has a channel type. Which you can use to determine which API you should use depending on the where it is heard from 😅
We have a slash command for /purpose which shows the purpose of the channel and we need to determine if it is heard from channel or group so we can use the bot.api.group.info or bot.api.channel.info accordingly

PR #1481, opened 2018-10-17
Move message delay functionality from tick function to message creation

ByrenHiggin:
The 'delay' property is not considered until the 'tick' function is called, I moved the timestamp assignment to the message creation to add delay functionality.

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -520,6 +520,10 @@ function Botkit(configuration) {
             } else {
                 message.channel = this.source_message.channel;
             }
+            if (message.hasOwnProperty(delay)) {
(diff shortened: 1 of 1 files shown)


benbrown:
Thanks for your contribution, @ByrenHiggin, but this isn't quite right.

The timestamp is being set on the _next outgoing message_  at the time the _current outgoing message_ is sent.  That way, the delay is offset from when the message is sent, not when it is enqueued.

Your change would cause issues for conversations created but not activated until later.  It would also not include the appropriate offset from the send time.

benbrown:
Fixed in #1484! Thanks.

PR #1496, opened 2018-10-27
basic botkit implementation of https://github.com/cdbhnd/botkit-viber

maaraanas:
need alot of cleanup and also merge it into one piece, but basic use is ok here)


examples/viber_bot.js
```diff
@@ -0,0 +1,50 @@
...
+                    default: true,
+                    callback: function (response, convo) {
+                        convo.gotoThread('default');
+                    },
+                }
+            ], {}, 'default');
+
+            convo.activate();
+        });
+    });
```
mokeddes: Return to the line (Newline at the end of file )

mokeddes: Recommanded to use arrow function

lib/ViberMessageTypes.js
```diff
@@ -0,0 +1,158 @@
...
+        'media': mediaUrl,
+    };
+
+    if (thumbnailUrl) { message[thumbnail] = thumbnailUrl; };
+
+    return message;
+}
+
+function VideoMessage(mediaUrl, size, duration, thumbnail) {
+    let message = {
```
mokeddes: Why not factor Message ? Like This:
let message = {
'receiver': '',
        'min_api_version': 1,
        'sender': {
            'name': '',
            'avatar': ''
        }
}

And you will add some proprieties for the message in all function

package.json
```diff
@@ -24,6 +24,7 @@
     "mustache": "^2.3.0",
     "promise": "^8.0.0",
     "request": "^2.81.0",
+    "request-promise": "^4.2.2",
```
mokeddes: Do you have see node-fetch or axios ?

lib/ViberMessageTypes.js
maaraanas: thanks! i'll fix this asap. Had no time in past few days at all

PR #1499, opened 2018-10-29
Webex guest

akalsey:
Webex recently introduced support for Persistent Guest Users. These are application-managed user identities, that instead of requiring a Webex Teams account to use the APIs, the application uses an existing user identity to generate a JWT and exchange it for an Oauth token. That oauth token can then be used to converse with other Teams users, both with messaging and with calls. Guests can't talk to other guests and have the same restrictions as free Teams accounts.

This PR adds guest support to Botkit. A developer can define the values that would make up their JWT, and then botkit would handle the token exchange and authenticate  to Teams. Because the oauth tokens are short-lived, the example bot also demonstrates how to refresh the token without restarting the bot.

### DIFF
--- a/domain-validate.patch
+++ b/domain-validate.patch
@@ -1,0 +1,35 @@
+diff --git a/lib/CiscoSparkbot.js b/lib/CiscoSparkbot.js
+index 6b6b708..883a688 100644
+--- a/lib/CiscoSparkbot.js
++++ b/lib/CiscoSparkbot.js
+@@ -157,13 +157,16 @@ function Sparkbot(configuration) {
+                 domains = controller.config.limit_to_domain;
+             }
+ 
++            address = require("email-addresses")
++            var a = address.parseOneAddress(message.raw_message.data.personEmail);
++
+             var allowed = false;
+             for (var d = 0; d < domains.length; d++) {
+-                if (message.user.toLowerCase().indexOf(domains[d]) >= 0) {
++                if (a.domain.toLowerCase() == domains[d]) {
+                     allowed = true;
+                 }
+             }
+-
++            
+             if (!allowed) {
+                 console.log('WARNING: this message came from a domain that is outside of the allowed list', controller.config.limit_to_domain);
+                 // this message came from a domain that is outside of the allowed list.
+diff --git a/package.json b/package.json
+index 3b59d98..b18fe69 100644
+--- a/package.json
++++ b/package.json
+@@ -14,6 +14,7 @@
+     "clone": "2.1.1",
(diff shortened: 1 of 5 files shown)


ObjectIsAdvantag:
hi @akalsey, before digging into the implementation, I am willing to understand more about the use case. I am curious why would we want Botkit to support "Permanent guest" accounts for the identity of Chat bots. Should we not stick to "Bot accounts" access tokens (long-lived / 100 years, and with Webex Teams integrated security for these: required @Mentions).
Could you elaborate. Thanks

benbrown:
I have some concerns about adding new dependencies to the main package that are only used for this edge case scenario.

I think this might make more sense as an external plugin rather than being part of the core library. It is already implemented _on top_ of the existing Webex adapter...

akalsey:
@ObjectIsAdvantag one use case is the web-chat support case. Customer fills out a form explaining their issue, then launches chat, which is implemented as guest. The issue description needs to be posted to the chat as the end user.

But more broadly, providing the tooling will expose the use cases as people innovate. Using these as psuedo bots 

@benbrown Agreed on the dependency side. Really wish NPM had some way of lazy dependency resolution. I can see if this can be pulled off with the existing dependencies, and if not, move the plugin route. Another thing that might be useful is to make this a part of the exising webex adapter instead of a different adapter.

ObjectIsAdvantag:
> one use case is the web-chat support case. Customer fills out a form explaining their issue, then launches chat, which is implemented as guest. The issue description needs to be posted to the chat as the end user.

What about the following implementation (this example is for Web)
- create an HTML form in a Web App (with customer authentication), 
- create a "Permanent Guest" account and transmit the temporary token (3 hours) valid to the HTML front-end
- initialize the Webex Teams SDK or Widget with that guest access token in the HTML front-end
- have the user interact from the Widget to the bot (implemented with a standard Botkit account)

Botkit would work as is (without need to support permanent guest).
Did I miss something? Please elaborate.

akalsey:
How does information from that web form end up inside the chat? A member of the chat must post it. Marcus needs help, so he fills out a web form asking to chat with an agent. The form asks for his name and his question. 

A guest ID gets generated, and a Widget is placed on a page as a chat between Marcus and Betty, the agent. We want to post Marcus's question into the space now, so that Betty knows what it was, and Marcus doesn't have to re-type it.

So we automate it from the Guest API, posting it as Marcus into the chat.

ObjectIsAdvantag:
Got it: it's not a bot interacting with the "Guest" user but an agent.
What's not clear to me is where Botkit can help to implement this use case ?

PR #1501, opened 2018-10-30
Allow conversation typing indicator using configuration

MarcL:
## Changes

- Allow typing indicator to be used in conversations by setting `allow_conversation_typing` to `true` in bot configuration when initialising.

## Context for raising PR

There have been previous discussions (below) as to how best to enable the typing indicator in conversations and no solutions have been agreed or merged. I'm suggesting a non-breaking configuration method which will allow people to turn it on as necessary when the bot is initialised but follows existing behaviour by default. Setting the `allow_conversation_typing` configuration to `true` will enable this. Feel free to suggest a different configuration property name. I've confirmed this works as expected when enabling a Facebook Messenger integration.

## Additional Info

I haven't added any tests as none currently exist for `CoreBot`. I'm happy to do so but wanted to confirm you were happy with this approach before starting. I also feel it might be tricky to test this functionality without a refactor of some of the existing functions. I didn't want to start refactoring in my first PR! 😄I don't think all platforms support `replyWithTyping` so I've guarded against this not existing too.

## Related issues / PRs
https://github.com/howdyai/botkit/issues/674
https://github.com/howdyai/botkit/pull/710

### DIFF
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -945,7 +945,12 @@ function Botkit(configuration) {
                                     that.sent.push(outbound);
                                     that.transcript.push(outbound);
 
-                                    this.task.bot.reply(this.source_message, outbound, function(err, sent_message) {
+                                    const {allow_conversation_typing} = this.task.bot.botkit.config;
+                                    const can_use_typing = this.task.bot.replyWithTyping;
+
+                                    const replyFunction = (allow_conversation_typing && can_use_typing) ? this.task.bot.replyWithTyping : this.task.bot.reply;
+
+                                    replyFunction(this.source_message, outbound, function(err, sent_message) {
                                         if (err) {
                                             botkit.log('An error occurred while sending a message: ', err);
 
--- a/lib/CoreBot.js
+++ b/lib/CoreBot.js
@@ -948,9 +948,9 @@ function Botkit(configuration) {
                                     const {allow_conversation_typing} = this.task.bot.botkit.config;
                                     const can_use_typing = this.task.bot.replyWithTyping;
 
-                                    const replyFunction = (allow_conversation_typing && can_use_typing) ? this.task.bot.replyWithTyping : this.task.bot.reply;
+                                    const reply_function = (allow_conversation_typing && can_use_typing) ? this.task.bot.replyWithTyping : this.task.bot.reply;
 
-                                    replyFunction(this.source_message, outbound, function(err, sent_message) {
+                                    reply_function(this.source_message, outbound, function(err, sent_message) {
                                         if (err) {
                                             botkit.log('An error occurred while sending a message: ', err);
 


MarcL:
I've committed a change to make the function name `reply_function` but it seems you use a mix of snake case (variables) and camel case (functions/methods). Is there a code styleguide so I can match it? Thanks.

MarcL:
Any interest in adding this? Would be useful for a current Messenger bot I'm working on.

benbrown:
Hey @MarcL - our policy on the typing indicator is that it should be applied using middleware on the send endpoint instead of being added into core. This is primarily because typing indicators are not supported by all the platforms.

MarcL:
Thanks for responding @benbrown. Makes sense to enable it at the platform adapter rather than in the core. Happy to close this PR but a few questions for the correct implementation.

The issue for me is that the `Conversation` calls the per platform `bot.reply` yet the platforms that support typing, also have an explicit `bot.replyWithTyping`.

If I add a middleware and append to the `send` chain that checks the config for an `allow_typing` flag, then this will be overriding all messages, even if it's explicitly sent with `reply` rather than `replyWithTyping`. Is this the expected behaviour in your mind?

I'd also have to add a `setTimeout` to get it to send the message with a delay. It feels like the middleware is the wrong place for this.

Happy to do the work on this but want to make sure I get this done in the place you expect.

Reading through the [FB sender action docs](https://developers.facebook.com/docs/messenger-platform/send-messages/sender-actions/), it does suggest responding to messages that a bot should respond to with the typing indicator but I don't want to change implied behaviour from the existing API.

Any thoughts?

MarcL:
Spoken to @benbrown in the Bot Developer Hangout. Will create a middleware to do this instead.

constantx:
> Spoken to @benbrown in the Bot Developer Hangout. Will create a middleware to do this instead.

any more details on this?

constantx:
I found https://github.com/MarcL/botkit-middleware-typing but seems to be unfinished :(

MarcL:
Hey @constantx! 👋 

I have implemented this locally but it's Facebook specific.
It would be useful if BotKit exposed some more generic functionality for whether an adaptor supports typing or not.

I will update my middleware in the next few days for you and then publish it with the FB specifics. I'm currently using it in my bot and it works well.

Thanks,
Marc

constantx:
Thanks @MarcL ! Greatly appreciated :) Let me know where I can buy you a 🍺

MarcL:
@constantx - I'll do this for you over the weekend. 👍

MarcL:
@constantx - I've just published v1.0.0 which supports Facebook typing only. Will look at Slack at some point soon - not sure if you need this. Please raise any issues if you spot any problems. Thanks!

https://github.com/MarcL/botkit-middleware-typing

constantx:
thanks @MarcL will use it and let you know :)

PR #1510, opened 2018-11-13
adding clientVerificationToken to SlackConfiguration

### LINKED ISSUE #1509: Typescript Definitions missing client verification token
Version: "^0.6.20"

A warning is given by TSC about clientVerificationToken if it is missing and in use when setting up the Slackbot Controller. This field is optionally used by `botkit/lib/SlackBot.js` but it is not defined for typescript.

The following would need to be added to the `SlackConfiguration` typescript definition
`clientVerificationToken?: string`

Will aim to submit a PR.

pellicceama:
Fixes #1509


benbrown:
Thank you!

PR #1536, opened 2018-12-11
Custom error message for cli

kobarasukimaro:
How to use 'say' function is wrong.
When building an environment with docker etc, it is difficult to understand the error message in an environment where git is not installed.


benbrown:
thanks!

PR #1537, opened 2018-12-11
Improve studio_token argument for cli

kobarasukimaro:
studio_token argument is optional. But it will be prompted to enter.
It is inconvenient to build botkit environment with Docker and I'd like to improve it.


benbrown:
This parameter is being removed from the next release of the cli.

PR #1539, opened 2018-12-11
use hmac compare instead of direct compare

danhofer:
As per https://api.slack.com/docs/verifying-requests-from-slack - "Compare the resulting signature to the header on the request. Note: for best practice, use an hmac compare function instead of directly comparing the signatures for equality."


benbrown:
Thanks, @danhofer

PR #1548, opened 2018-12-18
Change 500 to 404 when team not found

krashidov:
- We recently had an Enterprise Grid customer sign up. Somehow, we're receiving webhook events from teams that are part of their grid that haven't signed up yet and thus are not in our database. This has blown up our errors and now we're getting hundreds of emails a day.
- Please note that I originally changed these from simple return statements (that would eventually timeout) to 500 status codes. So I'm just doing what I should've done in the first place :)

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -285,7 +285,7 @@ function Slackbot(configuration) {
                 if (!slack_botkit.config.clientId) {
                     bot = slack_botkit.spawn({});
                 } else {
-                    return res.status(500).send();
+                    return res.status(404).send();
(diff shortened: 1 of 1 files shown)


benbrown:
Does this have an effect on Slack?  Does Slack stop sending the messages for that team?

krashidov:
No, it's more of this is the correct code we should be sending. We're on heroku and when we send 500's it's sending us emails thinking our server is on fire. When really it's just not finding a team.

benbrown:
Thanks! Merged, and will be in 0.7.1

PR #1558, opened 2019-01-10
adding  regexp to default ears

ab320012:
figured i would leave it here, seems like a default thing and i ended up having to go through source and a few of the docs to just to come to it


benbrown:
What is the goal here?

ab320012:
@benbrown uh just to be able to see what regexp passed instead of the match

PR #1579, opened 2019-01-30
slack api - extend error object with response details if rate limit is hit

ihorrusinko:
We recently got rate a limit error. An error message provides no information about  [`Retry-After`](https://api.slack.com/docs/rate-limits) time. This PR includes more response details when the rate limit error is hit.


benbrown:
Thanks!

PR #1594, opened 2019-02-21
slack api - support of blocks

ihorrusinko:
We tried to use new feature of  slack - `blocks` (in `chat.postMessage`) but botkit implementation does not support it. So we added `blocks`  to `format` middleware.


benbrown:
Thanks!

makstaks:
@benbrown or @ihorrusinko does the blocks option need to be sanitized? See: https://github.com/howdyai/botkit/blob/master/lib/Slack_web_api.js#L226

ihorrusinko:
@makstaks good catch, i will add update for `sanitizeOptions`

PR #1608, opened 2019-03-18
Add dialog state to Slackbot worker createDialog

ryansydnor:
To support: https://api.slack.com/dialogs#state


benbrown:
Thanks!

PR #1614, opened 2019-04-04
Bot mention args changes in middleware example

Luckyfoxdesign:


benbrown:
Both ways of specifying a list of events is acceptable. Why make this change?

Luckyfoxdesign:
@benbrown Because the methods in all slack examples where the types of bot mentions were not working until I made it an array.

PR #1636, opened 2019-05-07
Fixed misplaced ")" in facebook adapter init function...

colsen1991:
...causing the verify webhook to never get called.


colsen1991:
I've never used TS before, and haven't contributed with many PRs on github before. So please let me know if any more changes, tests etc. are needed.

I noticed this when trying to register verify webhook via developers.facebook.com. Seems the misplaced parenthesis caused the route to never register the handler as it was given as a second parameter to the getConfig-method. As such `GET webhook_uri` returned 404. This fix worked for me locally. Placing the parenthesis correctly and changing the callback into an arrow function in order for `this` to reference the class and not the function.

benbrown:
@colsen1991 thank you for the contribution. I will publish an update to this package this afternoon.

benbrown:
Published!

https://www.npmjs.com/package/botbuilder-adapter-facebook/v/1.0.2

colsen1991:
Cheers! Glad I could help :)

PR #1642, opened 2019-05-08
[Slack] Add redirectUri in getInstallLink()

jebarjonet:
On Slack adapter, `redirectUri` is not send to Slack via `getInstallLink()`. Therefore Slack is redirecting to a wrong url during installation, hitting the wrong endpoint and breaking installation.

If you use the Slack adapter with a `redirectUri` param before this fix, you cannot trigger OAuth with Botkit `getInstallLink()`.

### DIFF
--- a/packages/botbuilder-adapter-slack/src/slack_adapter.ts
+++ b/packages/botbuilder-adapter-slack/src/slack_adapter.ts
@@ -224,7 +224,7 @@ export class SlackAdapter extends BotAdapter {
      */
     public getInstallLink(): string {
         if (this.options.clientId && this.options.scopes) {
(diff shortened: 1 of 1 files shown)


benbrown:
Thank you for this - if you have time, could you make this a conditional addition to the URL if the configuration is set?

Can you also wrap it in encodeURIComponent?

See the legacy version:

https://github.com/howdyai/botkit/blob/legacy/lib/SlackBot.js#L637

jebarjonet:
@benbrown I updated the code accordingly

benbrown:
Thanks @jebarjonet! 

Are you a member of community.botkit.ai? If not, you should join us :)

benbrown:
@jebarjonet this is now in 1.0.2 on npm

PR #1653, opened 2019-05-10
fix(dialog): broader types for options list

roger-king:
Adding broader types for optionsList.


roger-king:
@benbrown Sorry I know I opened this PR before but after using it a bit more I found more use cases.

roger-king:
Just Note here that after testing different edge cases - I purposely left out `boolean` type. Slack will fail with a validation error if you use a `boolean` type inside of the `value` (default value)

roger-king:
@benbrown 

Sorry for the delay on this. I just updated the branch. Do you mind merging this in? 😄

PR #1667, opened 2019-05-20
Fix webex adapter to support FormData

Teamop:
1. the webex message api also supports to send messages with attachement, at this time, the request [`content-type` is `multipart/form-data`](https://developer.webex.com/docs/api/basics/message-attachments), so here the null value will break the [FormData used by request](https://github.com/form-data/form-data/blob/master/lib/form_data.js#L226).
2. `roomId`, `toPersonId` and `toPersonEmail` should be provided only one, otherwise, `BadRequest: Message destination could not be determined. Provide only one destination in the roomId, toPersonEmail, or toPersonId field`


Teamop:
@benbrown When it will be merged and released? Cause it blocks our development of the webex bot. Thanks

benbrown:
This will be part of 4.5!

PR #1668, opened 2019-05-20
WebAdapter: upgrade ws to 7.0.0

Naktibalda:
* Dropped support for Node.js 6
* Dropped support for url.Url instances in the WebSocket constructor
* The behavior of WebSocket#{p{i,o}ng,send}() has changed when the
readyState attribute is not OPEN
https://github.com/websockets/ws/releases/tag/7.0.0


benbrown:
I updated to 7.1.1

PR #1675, opened 2019-05-31
Change cms_uri to uri

adantoscano:
Botkit-cms-plugin expects uri instead of cms_uri


benbrown:
this will be part of 4.5! thanks.

PR #1677, opened 2019-05-31
Create uri with URL object

adantoscano:
Now you can write cms_uri with or whitout last /


benbrown:
This will be part of 4.5! thanks.

Issue #1681, opened 2019-06-07
Legacy Facebook adapter should be updated to use bot-specific config


benbrown:
One of the major limitations of the legacy Facebook adapters it is lack of support for multi-tenant bots.

In large part, this is because the spawned `bot` objects from this adapter use a adapter-wide global configuration option (`configuration.access_token` being the primary culprit) instead of a local config option (`bot.config.access_token`)

This would require a small change to the `defineBot` factory function to mix the global vars into the local config.

wolfejw86:
possible fix in https://github.com/howdyai/botkit/pull/1683

PR #1683, opened 2019-06-07
Fixes #1682 and #1681 [WIP]

### LINKED ISSUE #1682: Legacy Facebook adapter should have configurable API version
The api version is set in an internal constant, but should be settable (and possibly updated to LTS version of the API)

wolfejw86:
initial working support for multi-tenant facebook pages into legacy botkit plus possible strategy of scoping APIs to multitenant .spawn with a HOF that accepts which configuration to use and creates the apis object as a return.

### DIFF
--- a/lib/Facebook.js
+++ b/lib/Facebook.js
@@ -21,19 +21,19 @@ function Facebookbot(configuration) {
      * allowing them to be heard without subscribing to additional events
      */
     if (facebook_botkit.config.receive_via_postback) {
-        facebook_botkit.on('facebook_postback', function(bot, message) {
+        facebook_botkit.on('facebook_postback', function (bot, message) {
             facebook_botkit.trigger('message_received', [bot, message]);
         });
     }
 
     facebook_botkit.excludeFromConversations(['message_delivered', 'message_echo', 'message_read']);
 
-    facebook_botkit.middleware.format.use(function(bot, message, platform_message, next) {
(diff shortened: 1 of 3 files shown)


lib/Facebook.js
```diff
@@ -1876,6 +1876,21 @@ function Facebookbot(configuration) {
         }
     }
 
+    function scopeAPIsWithRuntimeConfiguration(configuration) {
```
wolfejw86: i put this here as one possible idea for how to allow for passing a spawn config to the api methods to allow for runtime page-access-token configuration.  i'm totally open to other ways of solving this, it was just my first thought for having to change the least amt of code

wolfejw86: looking at this a bit more, it looks like this could easily work for associating a config with all the api methods, however, tracing the path of where we have the `.defineBot` happening, i believe it's the global `botkit` facebook object getting passed in as the first param of the method passed in to `.defineBot`. Because of this, reassigning `bot.botkit.api = scopAPIsWithRuntimConfiguration(bot.config)` would have an unintended side affect of modifying it across a shared reference. 
not sure if it's worth assigning `bot.scopedApi = scopeAPIsWithRuntimeConfiguration(bot.config)`. while it would add a second api field on the spawned bot worker, it would also prevent any weirdness with mutating a global reference.  I don't have any other ideas at the moment, but will keep thinking on it. lmk what you think 😄

PR #1707, opened 2019-07-03
Don't add activity and reference fields to outgoing message

Naktibalda:
I noticed that Botkit adds some garbage fields to outgoing fields, e.g. `{"type":"message","text":"Hi...","channelId":"websocket","conversation":{"id":"98f515de-b10a-4b4c-51b9-d6fa3f0e1688"},"from":{"id":"bot"},"recipient":{"id":"98f515de-b10a-4b4c-51b9-d6fa3f0e1688"}}`

So I decided to remove them.

Previous activity check was incorrect and allowed falsy values to be added to channelData.
Reference fields weren't checked at all so they were sent to client.

### DIFF
--- a/packages/botkit/src/botworker.ts
+++ b/packages/botkit/src/botworker.ts
@@ -300,45 +300,55 @@ export class BotWorker {
      * @returns a properly formed Activity object
      */
     public ensureMessageFormat(message: Partial<BotkitMessage> | string): Partial<Activity> {
-        let activity: Partial<Activity> = {};
 
         if (typeof (message) === 'string') {
-            activity = {
+            return {
                 type: 'message',
                 text: message,
                 channelData: {}
             };
-        } else {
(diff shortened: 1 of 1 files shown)


Naktibalda:
Actually, I don't like this patch anymore,
I think that `applyConversationReference` call
at https://github.com/howdyai/botkit/blob/master/packages/botkit/src/botworker.ts#L136 is completely unnecessary, because the next step strips those fields.

Naktibalda:
I copied the good part of this change to #1708

Naktibalda:
Another U-turn,  I realized that this change was good.

benbrown:
This fix has been indirectly incorporated into 4.5! Thanks for the contribution!

Naktibalda:
Thanks, I upgraded to 4.5 and it is working as expected.

PR #1708, opened 2019-07-03
ensureMessageFormat: fix activity prop check

Naktibalda:
Previous condition allowed copying of properties having `undefined` value and potentially properties containing other falsy values.


Naktibalda:
#1707 is good.

PR #1716, opened 2019-07-08
Change conversations key to optional

adantoscano:
Convo.ask() without storage key throws error, It forces to add a empty object or caracter

### DIFF
--- a/packages/botkit/src/conversation.ts
+++ b/packages/botkit/src/conversation.ts
@@ -44,7 +44,7 @@ interface BotkitMessageTemplate {
     attachments?: any[];
     channelData?: any;
     collect: {
-        key: string;
+        key?: string;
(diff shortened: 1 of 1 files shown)


packages/botkit/src/conversation.ts
```diff
@@ -330,7 +330,7 @@ export class BotkitConversation<O extends object = {}> extends Dialog<O> {
      * @param key name of variable to store response in.
      */
     public ask(message: Partial<BotkitMessageTemplate> | string, handlers: BotkitConvoTrigger | BotkitConvoTrigger[], key: {key: string} | string): BotkitConversation {
-        this.addQuestion(message, handlers, key, 'default');
+        this.addQuestion(message, handlers, key = null, 'default');
```
benbrown: this part right here causes the key to always be set to null

Issue #1702, opened 2019-06-26
Filter received convo messages in middlewares


adantoscano:
### Are you sure this is an issue with the Botkit core module?
Yes

### What are you trying to achieve or the steps to reproduce?

I use `botkit-middleware-dialogflow` in ingest middleware, but I don't need to send to dialog messages that responds to a convo question. At which part of the middleware can I realize that the message is to answer to a convo?

### Context:

- Botkit version: 4.0.2
- Messaging Platform: Web
- Node version: 10.15.0
- Os: Linux
- Any other relevant information: Botkit FTW

benbrown:
This will be achieved by the new `interpret` middleware that fires AFTER dialogs are continued, but BEFORE any triggers or events are fired.

PR #1717, opened 2019-07-09
Fix middleware pipeline

adantoscano:
Receive middleware is not running when it match with a hear() trigger.

I moved hear check into receive middleware

#1702


benbrown:
great job @adantoscano thanks for tackling this.

benbrown:
FYI @adantoscano I am modifying your work to move receive even further up in the process, right after ingest, so that receive will fire BEFORE dialogs are consulted.

I am also introducing a NEW middleware that will fire between dialogs and events - this is now called "interpret."

Issue #1710, opened 2019-07-04
Convo.ask only gets text field as response


adantoscano:
### Are you sure this is an issue with the Botkit core module?
Maybe with BotFramework Dialogs

### What are you trying to achieve or the steps to reproduce?
I want to use other data in convo.ask rather than text

````
 convo.ask({
    text: 'Sí o no',
    channelData: {
      quick_replies: [{
        text: 'sí',
        payload: 'yes'
      }, {
        text: 'no',
        payload: 'no'
      }]
    }
  }, async (resp, convo, bot) => {
    console.log('RESP: ', resp.payload); //error resp is a string
  }, 'coso');
````

### What was the result you received?
just a string

### What did you expect?
Get the complete message

### Context:

- Botkit version: 4.0.2
- Messaging Platform: Any
- Node version: 10.15.0
- Os: Linux
- Any other relevant information:

**Comment by naikus** [comment]
@adantoscano 
This is a blocker for us. In previous version, we had a RASA NLU middleware that identified intents and we did something on these lines:
```javascript
convo.addQuestion("#AMA", (res, convo) {
   const {intent, entities} = res;
   convo.setVar("intent", intent);
   convo.setVar("entities", entities);
   convo.gotoThread(intent || "dontunderstand");
});
```

Now there is no way to do this :(

adantoscano:
Feel free to review my pull req #1719. I found that problem is on Prompt because this is to get text only, but I don know how to send the whole message through.

naikus:
Hi @adantoscano I locally did those changes but that did not work. Essentially the message that goes through the controller middleware is not sent to the handlers anywhere to the dialogs. See [L746](https://github.com/howdyai/botkit/blob/master/packages/botkit/src/core.ts#L746) and then [L755](https://github.com/howdyai/botkit/blob/master/packages/botkit/src/core.ts#L755) 
Here the message object created on [717](https://github.com/howdyai/botkit/blob/master/packages/botkit/src/core.ts#L717) but never gets further to dialogs :-(

Do you think this implementation is broken? botkit v0.7 allowed middlewares to modify message objects in a lot of ways. Here I see no way of doing this.

[...]

PR #1719, opened 2019-07-10
Send whole activity through resume dialog

adantoscano:
With this approach you can use other message params to match paths and use it #1710


adantoscano:
It does not work because `dialogContext.continueDialog()` is not calling to the changed one.

adantoscano:
I found the problem is on Prompt used(TextPrompt) it only gets text from activity. I tried to use ActivityPrompt instead but I need to pass a validator.

Any Idea?

Issue #1701, opened 2019-06-26
Convo messages don't go through send middleware


adantoscano:
### What are you trying to achieve or the steps to reproduce?

I want to log all messages passing through botkit, but conversation messages don't go through send middleware. How can I get it?

### What was the result you received?
none

### What did you expect?
conversation message info

### Context:

- Botkit version: 4.0.2
- Messaging Platform: Web
- Node version: 10.15.0
- Os: Linux
- Any other relevant information: Botkit FTW

benbrown:
Right now to do this you will need to use the botbuilder middleware layer, which is a bit different than Botkit:

https://botkit.ai/docs/v4/core.html#botbuilder-adapter-middleware

You can see all incoming AND outgoing activities with this middleware.

To get outgoing, you have to attach an event handler to the turn context inside the middleware - I know, it is a weird pattern.

You can see this in action within the TranscriptLogger class that is included in `botbuilder`.  You may want to just use this!

https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder-core/src/transcriptLogger.ts

benbrown:
HOWEVER this is a known issue and I do plan to address it.

benbrown:
Fixed in 4.5

PR #1720, opened 2019-07-10
Fix convo messages don't go through send middleware

adantoscano:
Related to #1701


benbrown:
Good effort, but this is missing a bunch of promise handling which will cause this to break.  I am going to attempt a fix like this myself.

PR #1725, opened 2019-07-15
fix output text is truncated

Teamop:
Currently, when using the `addMessage` to define a thread, like this:
```
convo.addMessage({
    text: 'Anyways, moving on...',
    action: 'next_step',
});
```
Then when changing to this thread later, the text is truncated to some random character, as for the implementation -- 
`outgoing = MessageFactory.text(line.text ? line.text[Math.floor(Math.random() * line.text.length)] : '');`

So here, make a change for the text to always be an array to be the same with other implementation, so that can just get the first item from the array instead of some random character.


benbrown:
This will be fixed in the next release.

Issue #1094, opened 2017-10-16
[help] How use mustache variables as an array, dynamically?


gcfabri:
I store inline messages and structured messages in separate files from business logic.
Out of message files, when I set a variable like this:

bot.js:
```javascript
//...
convo.setVar(`sample`, `foobar`);
//...
```
and then make a reference in this file inline_messages.js:

```javascript
export default {
  myMessage: `This is my variable: {{ vars.sample }}`
}
```
and use it on any conversation, it works nicely.
`This is my variable: foobar`

How can I use this as an array to build a structured message dynamically (eg.: Facebook Generic Template)? There is a way to do it using Mustache lib or Botkit?

Please, take a look:

bot.js:

```javascript
convo.setVar(`sample`, [{ cod: '1', value: 'foo' },{ cod: '2', value: 'bar' }]);
```

structured_messages.js:

```javascript
export default {
  myStructuredMessage: {
    type: `message`,
    text: ``,
    attachments: [ 
      // How to iterate this array saved on variable 'sample' and use it to build this structure dynamically? 
    ]
  }
}
```

benbrown:
There is not currently a way to dynamically generate attachments based on variables.

This is definitely something we're interested in adding, but I haven't decided on how it should work yet. Any thoughts?

akhanukov:
Before I knew what Mustache Script was, and/or when I need global variables, I'm usually pulling from a database (via API) of messages tied to a common topic, flow, conversation, or what we call "interactions", and build an array of those messages with keys as message IDs as @benbrown taught me. Those are of course navigated between using changeTopic or whatever but that's neither here nor there.

However, the places where we need additional variables, within the "output messages array", we simply do a find and replace where we use something like a WordPress shortcode notation, ie. `[variable1]` or in some cases something made up like `-_-` , natch. ;) So basically, doing what Mustache is created to alleviate, I think.

See, the problem I have, is: if build a wealth of Conversations (Topics, Flows, `convo`s basically), and in addition to jumping around within that topic or conversation tree, I want to jump between topics, I write my own filter in the `on` or `hears` handler which works. However, that creates a new tree and any `convo` variables go bye bye. I thought of using server variables like how I store my Customer ID in the command to run my bot.js file, but that seems... wrong?  I get why I don't want to store sensitive or even temporary variables in a MySQL nor even a NoSQL database, but what's the best way to do both what @gcfabri and/or I need? (If this is a separate issue, let me know and I'll create a ticket to discuss.) Thanks!

PR #1731, opened 2019-07-22
enable mustache rendering of quick_replies button

InnoraG:
This pull request simply enable missing mustache template rendering of quick_replies button in addition to text, it can add dynamic flexibility as response to kind of issue [#1094]

Result is visually explained from Emulator as below :

Without the change :

image

With the change :

image


benbrown:
This will be in 4.5! Thank you.

PR #1743, opened 2019-07-25
Update core.md re: spawn()

roeintense:
Our team found out that you can simply pass in the string of the team_id that you want to spawn the bot for, which is actually _easier_ than the previous version, but the documentation is super confusing with its reference to dialogContext and turnContext. Hope this or something similar helps.

### DIFF
--- a/packages/docs/reference/core.md
+++ b/packages/docs/reference/core.md
@@ -456,7 +456,7 @@ for handling platform-specific events or activities.
 
 | Argument | Type | description
 |--- |--- |---
-| config (optional)| any | Preferably receives a DialogContext, though can also receive a TurnContext. If excluded, must call `bot.changeContext(reference)` before calling any other method.<br/>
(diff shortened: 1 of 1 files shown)


benbrown:
Agreed that the core reference doc needs to point out the platform-specific behaviors.

I think the comment should be more platform-agnostic, or at least link to the other variations.

benbrown:
oops didn't mean to close that.  Also thank you for pointing this out!

packages/docs/reference/core.md
```diff
@@ -456,7 +456,7 @@ for handling platform-specific events or activities.
 
 | Argument | Type | description
 |--- |--- |---
-| config (optional)| any | Preferably receives a DialogContext, though can also receive a TurnContext. If excluded, must call `bot.changeContext(reference)` before calling any other method.<br/>
+| config (optional)| any | Preferably receives a DialogContext, though can also receive a TurnContext. If excluded, must call `bot.changeContext(reference)` before calling any other method.<br/>Can also take the string of the Slack ID of the team for which you'd like to spawn the bot. See <a href='https://github.com/howdyai/botkit/tree/master/packages/botbuilder-adapter-slack for more information'>the botbuilder Slack adapter readme</a> for details.
```
benbrown: There is a specific place in the Botkit docs that talk about spawning with a team id:

https://botkit.ai/docs/v4/reference/slack.html#create-a-new-slackbotworker

Other platforms have a similar note.

PR #1744, opened 2019-07-25
Fix typo breaking plugin middleware example code

ron-reaser:
The example code for the form of a plugin with automatically bound middlewares was incorrect. The actual key that usePlugin()'s source code refers to is `middlewares`, not `middleware`. Without this fix in the example code, middlewares are not automatically bound. With this fix, they are bound as expected.


benbrown:
Thanks, @ron-reaser!

PR #1746, opened 2019-07-26
Add attachment without 's' to make outgoing

adantoscano:


benbrown:
Good catch.  Should also probably add .blocks

benbrown:
Thanks!

Issue #947, opened 2017-08-04
Is there any integrate solution in botkit to build a Multilangual bot


MagSag:
I am building a facebook bot and would like it to be replying in different language according to the user localization or language parameter.

So I would like to know if there is any way to do that with botkit only and how?  if not which module would you suggest with an easy way of integration.

peterswimm:
There is currently no multilingual support in Botkit, but you could perhaps manage something using variables and threads where you:

- Start the conversation with asking a user to chose their language
- Store that choice
- Have all your scripts's text rendered as strings, and as users navigate through the conversations, you add your localized strings from Storage on the fly.

Obviously not ideal, we hope to support localization in the future, but no definite plans as of yet.

MagSag:
Thanks;

Issue #41, opened 2015-12-23
Add support for Internationalization/Localization


rodrigoalviani:

RafaelCosman:
@rodrigoalviani, great point. This is important.

peterswimm:
Created a new dev4slack channel, #botkit-localization to discuss this Feature Request. If you are interested, please join the conversation!

znat:
For now I am doing the following (with `i18n-node`):

At bot initialization:

```js
var i18n = require('i18n');
i18n.configure({...});
var bot = controller.spawn({...});
bot.i18n = i18n;
```
I wrote a [middleware](https://github.com/mrbot-ai/botkit-middleware-fbuser) to add facebook messenger's `user_profile` to the message object:
```js
{ text: 'ce soir',
  user: '123456789',
  channel: '/#QpmoHYNhZlvOzUXiAAAB', // socket id 
  timestamp: 1484084415264,
  user_profile: 
   { 
     first_name: 'Nathan',
     last_name: 'Zylbersztejn',
     locale: 'fr_CA',
     profile_pic: '<url>',
     timezone: -5,
     gender: 'male',
     is_payment_enabled: true,
     id: '123456789' },
  entities: {...} 
}
```

In the conversation:
```js
 bot.createConversation(message, function (err, convo) {
        convo.setVar("user_profile",message.user_profile);
        convo.addMessage({
            text: bot.i18n.__({phrase:'reminder_confirmation',locale:convo.vars.user_profile.locale}),
            action: 'completed'
        },'set_reminder');
       // rest of convo...
}
```

It would make sense to have the user profile accessible in the convo (which is also semantically defined by the people who run it). This would avoid setting the variable and we could find a way to make i18n aware of it to avoid passing the locale at every translation.

PR #1747, opened 2019-07-27
Support i18n via 'text' functions

cooperka:
## Motivation

Our team needed **per-user** internationalization, so I went with a quick and easy solution that works for us. If this deserves more thought before being integrated into the official codebase, I'm happy to wait on the PR and help do it "right". Just let me know what you need.

Related issues: #41, #947

## Changes

Adds support for `text` as a function, in addition to the existing string or array options, e.g.

```ts
// Simple string:
convo.say('Hello');

// Randomly picks one possibility from the array:
convo.say({ text: ['Hi', 'Hello'] });

// New function-based text:
convo.say({ text: (vars) => vars.isMorning ? 'Good morning' : 'Good evening' });
```

## Example

Code snippets below for a multilingual bot using the `polyglot` lib for translations (though other libs like `i18next` would work just as easily).

First set the user's locale on every message via middleware (perhaps you've saved their preference in your database):

```js
function setUserLocale(bot, message, next) {
  const locale = determineUserLocale(message.user);
  bot.getConfig('activity').userDeclaredLocale = locale;
  next();
}

controller.middleware.ingest.use(setUserLocale);
```

Then save the locale as a conversation variable at the beginning of the conversation:

```js
const convo = new BotkitConversation('myConvo', controller);

convo.before('default', (convo, bot) => {
  convo.setVar('locale', bot.getConfig('activity').userDeclaredLocale);
});
```

Finally, use whatever method you want to in order to translate each outgoing message:

```js
convo.say({ text: (vars) => i18n[vars.locale].t(text) });
```

It does indeed work with our fork -- just modify your package.json dependency to:

```json
"botkit": "sassafrastech/botkit#2019-07-28-textFn-v4.5.0-2"
```

Note: Full docs/testing not yet implemented, since there will probably be some discussion before accepting this PR. Again, I'm happy to make changes to allow wider use of this feature.


benbrown:
Interesting solution, well documented!

I'll have to ponder this a bit to make sure it doesn't break the world.

smoyth:
@benbrown Any updates here? Any support we can provide? This is going to be a core part of our use of botkit so if it's going to "break the world" we'll need to know soon :D

I see another person opened a similar PR: https://github.com/howdyai/botkit/pull/1799

Hopefully we can figure something out...

benbrown:
I am planning on including this in the next release!

smoyth:
Wow, great!

PR #1758, opened 2019-08-01
Fail with better error when thread not found

cooperka:
### Motivation

I was helping a friend debug a conversation earlier and the error we were getting didn't make any sense. Then we realized the thread name just had a typo in it.

Demo project with errors: [comet-spice on Glitch](https://glitch.com/edit/#!/comet-spice?path=features/sample_echo.js:12:0).

### Changes

Improves the following error message:

When trying to go to a thread that hasn't been defined, e.g. `convo.gotoThread('foo')`:
```console
UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'length' of undefined
    at BotkitConversation.<anonymous> (node_modules/botkit/lib/conversation.js:523:37)
```
=>
```console
UnhandledPromiseRejectionWarning: Error: Thread 'foo' not found, did you add any messages to it?
    at BotkitConversation.<anonymous> (node_modules/botkit/lib/conversation.js:466:19)
```


benbrown:
Good change, thanks!

PR #1761, opened 2019-08-07
correct mustache rendering from causing mutation of initial dialog script

InnoraG:
prevent `mustache.render` or `parseTemplatesRecursive` to mutate back `line` objects through copying to `outgoing` by value instead of reference using deepcopy format `JSON.parse(JSON.stringify(line.xxxxx))` of objets like attachments, quick_replies or blocks


InnoraG:
Adding example for better clarity : 
In a sample multilang bot using mustache variable as template for translation, without deepcopy, the mustache template variable gets rendered permanently and disappear in "line" variable preventing rendering it over again 

Sample rendering without deepcopy (QR text is fixed as initial value) :

NoDeepCopy2


Sample rendering with deepcopy (QR text can be translated using mustache in each turn) :

WithDeepCopy2

benbrown:
Thank you!!

PR #1763, opened 2019-08-08
added quick_reply to facebook_event_middleware

hukpavlo:
Very small changes for handling facebook quick_reply. I tested and it works with new version of botkit.

### DIFF
--- a/packages/botbuilder-adapter-facebook/src/facebook_event_middleware.ts
+++ b/packages/botbuilder-adapter-facebook/src/facebook_event_middleware.ts
@@ -53,6 +53,8 @@ export class FacebookEventTypeMiddleware extends MiddlewareSet {
                 type = 'facebook_postback';
             } else if (context.activity.channelData.referral) {
                 type = 'facebook_referral';
+            } else if (context.activity.channelData.quick_reply) {
+                type = 'facebook_quick_reply';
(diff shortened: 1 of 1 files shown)


benbrown:
Why do you want to classify it as a quick reply instead of just being treated as a normal message?

hukpavlo:
> Why do you want to classify it as a quick reply instead of just being treated as a normal message?

because i am writing facebook shop bot now without user input and i should use a function for handling this quick replies, i think it will be better writing only one word.
for example
controller.hears('SHOP', 'facebook_quick_reply', async (bot, message) => {});
controller.hears(message => message.quick_reply.payload === 'SHOP', 'message', async (bot, message) => {});

benbrown:
Are facebook quick replies different than the postback events?

hukpavlo:
> Are facebook quick replies different than the postback events?

Yes, facebook_postback doesn't handle facebook quick reply

benbrown:
OK.

In addition to setting the event type, you will also need to copy the quick reply payload into message.text.  That should happen around here:

https://github.com/howdyai/botkit/blob/master/packages/botbuilder-adapter-facebook/src/facebook_adapter.ts#L418

benbrown:
@hukpavlo did you find an alternate solutoin?

Issue #1767, opened 2019-08-12
facebook API call with url params


adantoscano:
### Are you sure this is an issue with the Botkit core module?
With Facebook adapter

### What are you trying to achieve or the steps to reproduce?
Get graph facebook info calling to /<fuid>?fields=locale

### What was the result you received?
`Syntax error "Expected end of string instead of "?"` trying ```await bot.api.callAPI(`/${psid}?fields=locale`, 'GET', {});```

### What did you expect?
Get locale info

### Context:

- Botkit version: 4.5.0
- Messaging Platform: Facebook
- Node version: 10
- Os: Linux
- Any other relevant information:

adantoscano:
merged into master

PR #1768, opened 2019-08-13
Fb api query params

adantoscano:
I've added query object to facebook api class to allow request with query param. I've created two new methods to make it more readable, `bot.api.callAPI()` is redundant, `bot.api.get(path, query)` and `bot.api.post(path, body, query)` use callAPI in a convenient way.

This is a way to fix my issue #1767


benbrown:
@adantoscano is it ever necessary to use the payload AND the query params? 

Instead of adding a new parameter and 2 new methods, what about just using the payload to hold query params when it is a GET?

adantoscano:
Done @benbrown

packages/botbuilder-adapter-facebook/src/facebook_api.ts
```diff
@@ -47,15 +47,26 @@ export class FacebookAPI {
...
      */
-    public async callAPI(path: string, method: string = 'POST', payload: any): Promise<any> {
-        let proof = this.getAppSecretProof(this.token, this.secret);
+    public async callAPI(path: string, method: string = 'POST', payload: any = {}): Promise<any> {
+        const proof = this.getAppSecretProof(this.token, this.secret);
+
+        let queryString = '?';
+        let body = {};
+
+        if (method === 'GET') {
```
benbrown: use toUpper here so the method is not case sensitive

packages/botbuilder-adapter-facebook/src/facebook_api.ts
```diff
@@ -47,15 +47,26 @@ export class FacebookAPI {
...
+            for(const key in payload) {
+                queryString = queryString + `${key}=${payload[key]}&`;
```
benbrown: should use encodeURIComponent for these

adantoscano:
Done!

PR #1777, opened 2019-08-21
Support .ts (or parameterized) file extensions for loadModules

Mutmatt:
A relatively small change that showed up when I was using typescript for my features in the demo bot.

### DIFF
--- a/packages/botkit/src/core.ts
+++ b/packages/botkit/src/core.ts
@@ -1125,7 +1125,7 @@ export class Botkit {
      * @param p {string} path to a folder of module files
      * @param exts {string[]} the extensions that you would like to load (default: ['.js', '.ts'])
      */
-    public loadModules(p: string, exts: string[] = ['.js', '.ts']): void {
+    public loadModules(p: string, exts: string[] = ['.js']): void {
         // load all the .js|.ts files from this path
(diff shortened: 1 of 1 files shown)


benbrown:
But the code that actually runs is the .js files. You won't ever be executing against the uncompiled ts files.

Mutmatt:
That wouldn't exactly be true if you are running `ts-node`. Yes, all the way down it is javascript but at a base level I wouldn't be checking in `.js` files therefor it wouldn't pick up my `/features/*.ts` files

benbrown:
I've never seen that!

Mutmatt:
`ts-node` 🔥! Welcome to the FUTURE! :D

packages/botkit/src/core.ts
```diff
@@ -1123,10 +1123,13 @@ export class Botkit {
      * ```
      *
      * @param p {string} path to a folder of module files
+     * @param exts {string[]} the extensions that you would like to load (default: ['.js', '.ts'])
```
Mutmatt:
```suggestion
     * @param exts {string[]} the extensions that you would like to load (default: ['.js'])
```

maximebeaudoin:
Yes please !

PR #1800, opened 2019-09-07
Include http server in botkit.webserver

ryansydnor:
We need the ability to call `server.close()` to stop our server (in our integration test suite) and the http server is not currently exposed. This fixes that!


ryansydnor:
Included in https://github.com/howdyai/botkit/pull/184. Closing.

PR #1801, opened 2019-09-08
Support entire message objects in convo handlers. Fixes #1793 and #1710

### LINKED ISSUE #1710: Convo.ask only gets text field as response

### LINKED ISSUE #1793: Incoming messages augmented by botkit middlewares are not available to conversation handlers
Yes

This is a blocker for us currently as we have migrated from 0.7 with some new features like multiple bots per platform (i.e. web, facebook, slack) and also migrated the middlewares to the plugin API of the new version (4.x)

We have a RASA middleware that augments the message with intent and entities. But these messages are not available to conversation handlers.

### Describe your issue here, include as much detail as necessary to reproduce the issue or implement a missing functionality.

We have a RASA Middleware like so
```javascript
controller.usePlugin(botkit => {
  return {
    name: "Rasa NLU",
    middlewares: {
      receive: [
        (bot, message, next) => {
          queryRasa(message.text).then(res => {
            message.entities = res.entities;
            message.intent = res.intent;
            next();
          });
        }
      ],
      send: []
    }
  };
})
```

```javascript
const convo = new BotkitConversation("smalltalk", controller)
convo.addQuestion("Let's chat", (res, convo) => {
   const {intent, entities} = res;
   convo.setVar("intent", intent);
   convo.setVar("entities", entities);
   convo.gotoThread(intent || "dontunderstand");
});
controller.addDialog(conversation);
```

I receive the text sent by the user (```message.text```)
```javascript
convo.addQuestion("Let's chat", (res, convo) => {
  console.log(res); // Hi! (If the user entered Hi!)
});
```

### What did you expect?
The handler for the question to get message with intent and entities augmented by the Rasa NLU middleware
```javascript
convo.addQuestion("Let's chat", (res, convo) => {
  console.log(res);
});
```
This should print:
```javascript
{
  text: "Hi!",
  intent: "greet", // Identified by Rasa NLU
  entities: []
  ...other message properties
}
```

### Context:

- Botkit version: 4.0x
- Any other relevant information: There is also a bug: #1710 which I think might be related to this, where I've commented some of my findings [here](https://github.com/howdyai/botkit/issues/1710#issuecomment-527897136)

naikus:
This merge request allows botkit conversation handlers (those that go with ```addQuestion``` api) to receive full message objets. ( Fixes #1710 )

Also any augmentation done by botkit middlewares to messages are now available in handlers. This is especially useful for NLU intent and entity identification (i.e. if any of the handlers need to know them) ( Fixes #1793 )


benbrown:
Thanks for this work.

I have some initial concerns about breaking changes - even things like changing text from optional to required could break things for devs.

Does this change the signature of any handlers?

naikus:
Hi @benbrown, thanks for replying. This does not change signatures for any existing handlers. The change is completely internal to botkit.

In fact existing conversation handlers (in botkit v0.7x) expected message objects which changed to only text. Also in case of questions only text can anyway be captured in capture key (as a convo var)

The BotkitMessage text was made a 'required' property because the class that it extends from has that set as required. 

Thanks,
A

benbrown:
I am on this and plan to get it reviewed and included in the next release if at all possible.

benbrown:
Hi @naikus,
After careful review of this I can't accept it in its current state.
The key issue is that changing the signature of the handler functions that currently receive a string value to receiving the entire message payload is a breaking change.

However, I do think this is a valid issue and you've done 90% of the work.

In order to merge this, I request the following change:

Instead of swapping the text field for the full message, add an additional parameter to the handler signature.
This would result in a function that receives `(answer, convo, bot, full_message)` 

This way, the change is additive and will not break existing bots.

naikus:
Hi @benbrown, I agree. I'll make the changes and to the signature and push the changes for review.
Thanks!

benbrown:
Thanks @naikus! Ping me when you push the new commits.

etiennellipse:
@naikus Any updates on this?

naikus:
> @naikus Any updates on this?

Hi @etiennellipse 
I'm a bit busy with some tasks. But I'll have this push in soon. (End of dec tentatively or earlier if possible)
Thanks,
Aniket

naikus:
Hi @benbrown,
I've pushed the commits according to the API you suggested.
The callback handlers for questions now take 4 arguments:
```javascript
convo.addQuestion(
    "what?",
    async (response, convo, bot, message) => {
      // This logs the entire message payload (BotkitMessage) where message.text is same as response
      console.log(message); 
      await convo.gotoThread('thread-3');
    }, 
   "what",
   "thread-2"
);
```

deepakmahakale:
I have been waiting for this change for months now :smile: 
Kudos

NxP4Code:
I am also waiting for this one for a long time .. hope this gets approved soon and comes in next release.

It may fix #1862

benbrown:
I'm on it!

naikus:
I've done some e2e testing with this. I was wondering if it'd help if there were unit tests for this. I can start on those. Only I did not see any existing tests for callbacks for ```convo.addQuestion```

benbrown:
@naikus yes, new tests would be wonderful.

naikus:
@benbrown I've add necessary unit tests, thanks.

sureshvarman:
+1

packages/botkit/src/core.ts
```diff
@@ -689,6 +694,35 @@ export class Botkit {
...
+            ...message,
+            // timestamp: new Date(),
+            type: message.type === 'message' ? ActivityTypes.Message : ActivityTypes.Event,
+            // normalize the user, text and channel info
+            event: channelData.type || channelData.botkitEventType || message.type,
+            text: message.text,
+            user: message.from.id,
+            channel: message.conversation.id,
+            value: message.value,
+            incoming_messge: message,
```
adam-resdiary: Should this not be `incoming_message`?

benbrown: Yes good catch I will probably not merge this piece anyways since it is sort of out of scope of the main change.

benbrown:
working on merging this now.

benbrown:
snag #1: the way this is implemented to create a new turncontext breaks a bunch of the expected adapter behavior regarding the turn state.  however, without it, we don't get access to the message object we want.  

experimenting!

benbrown:
phew! I was able to work around that issue by using the turnContext.turnState to store the botkit message and make it available inside the dialog.

benbrown:
OK! This functionality is included in the next release -> #1921

adam-resdiary:
@benbrown nice one! Any idea when the new release is likely to be made available?

benbrown:
@adam-resdiary this week unless a big snag is encountered on the botbuilder 4.8 release, which I am waiting for. Should be tomorrow!

benbrown:
AT LONG LAST, this has been merged!

pgoldweic:
@benbrown, does this work also for the convo.after handler? The documentation does not include a full_message parameter in there. I'm realizing this would resolve a legacy migration issue that I'm having ( in which I need the full message object once a -single question- dialog ends). In legacy botkit I was able to use a dynamic convo for this (created within a controller.hears() handler).

PR #1815, opened 2019-09-19
Botkit dialog test client

qwertyuu:
Hello! Our team is starting to love the library you are developing. Since we are starting to consider writing tests to make sure our botkit bots are behaving as expected.

Since we could not find a specific way to test the library, I went ahead and searched around the web for a way to test botbuilder bots. Microsoft seems to provide a way to do it: https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder-testing/src/dialogTestClient.ts

So I went ahead and brought that code back into Botkit's code base and adapted it to take a Botkit instance and a thread name to test. I have provided an example test file showing how to use the test client in the PR.

This will be immensely useful for both our team and the whole botkit community.

### DIFF
--- a/packages/botkit/package.json
+++ b/packages/botkit/package.json
@@ -10,7 +10,8 @@
   ],
   "scripts": {
     "build": "tsc",
-    "eslint": "./node_modules/.bin/eslint --fix src/*"
+    "eslint": "./node_modules/.bin/eslint --fix src/*",
+    "test": "tsc ; nyc mocha tests/*.tests.js"
   },
   "author": "benbrown@gmail.com",
   "license": "MIT",
--- a/packages/botkit/src/index.ts
+++ b/packages/botkit/src/index.ts
@@ -10,3 +10,4 @@ export * from './core';
 export * from './conversation';
 export * from './botworker';
 export * from './dialogWrapper';
+export * from './testClient';
--- a/packages/botkit/src/testClient.ts
+++ b/packages/botkit/src/testClient.ts
@@ -36,7 +36,7 @@ export class BotkitTestClient {
      * assert.strictEqual(reply.text, 'first reply', 'reply failed');
      * ```
      *
-     * @param channelOrAdapter The channelId to be used for the test or custom adapter. By default the test client will use a default TestAdapter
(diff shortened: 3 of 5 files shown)


qwertyuu:
By the way, I don't think `tests/Core.tests.ts` is being run when executing `npm test`. I might be mistaken though, I have no idea how lerna works

etiennellipse:
I got it running through lerna. I added the `test` script to botkit's package.json, and modified the core test so that it does not hang the process.

People can now add coverage when fixing issues on the dialog engine, which I believe is a great enhancement.

ianfixes:
Can you add some documentation for how this code can be leveraged to write tests of a bot?

etiennellipse:
@ianfixes Check the packages/botkit/tests/Dialog.tests.js file, that is the starting point. Actually, this addition allows a much better test coverage of the Botkit project itself in the future since it makes dialog testing possible in isolation.

ianfixes:
In this example:

```js
    it('should follow a dialog', async function () {
        const introDialog = new BotkitConversation('introduction', bot);
        introDialog.ask({
            text: 'You can say Ok',
            quick_replies: [{
                title: 'Ok',
                payload: 'Ok'
            }],
        }, [], 'continue');
        bot.addDialog(introDialog);

        // set up a test client
        const client = new BotkitTestClient('test', bot, 'introduction');

        // Get details for the reply
        const quickreply_reply = await client.sendActivity();
        assert(quickreply_reply.text === 'You can say Ok');
        assert(quickreply_reply.channelData.quick_replies[0].title === 'Ok');
    });
```

It looks like you're creating the bot immediately before testing it. In normal use, I'd be instantiating my own bot-under-test somehow, and the test would simply have a-priori knowledge that there was a skill called `'introduction'`... right?

qwertyuu:
@ianfixes Yes. The test I wrote is "self-contained" as an example. The dialog should not be known to the test, other than its name (`introduction` here) and its expected outcomes.

benbrown:
This is super awesome, btw.

I am using this testing tool to validate the 4.6 changes.

Hooray!

PR #1818, opened 2019-10-01
Fix ask/addQuestion type signatures

etiennellipse:
Developing with TypeScript, I stumbled upon issues with the ask & addQuestion signature. As expected by the `handlers` parameter, I passed a single `BotkitConvoTrigger` object instead of a function, which caused this error:
```
message.collect.options.forEach((o) => { if (!o.type) { o.type = 'string'; } });
                                ^
TypeError: Cannot read property 'forEach' of undefined
    at BotkitConversation.addQuestion (/Users/etienne/Code/cerebro/node_modules/botkit/src/conversation.ts:379:33)
    at BotkitConversation.ask (/Users/etienne/Code/cerebro/node_modules/botkit/src/conversation.ts:333:14)
```

Checking the code, it seems obvious that a single `BotkitConvoTrigger` is not expected; it is either a `BotkitConvoHandler`, or an array of `BotkitConvoTriggers`. Otherwise, the conditional block on lines 367:376 fails.

Here is an example of TypeScript code that validates correctly for the compiler, but crashes at runtime:
```
const { BotkitConversation } = require("botkit");

module.exports = function (controller) {
    const dialog = new BotkitConversation("intro", controller);

    dialog.ask({
        text: ["Need anything?"],
        quick_replies: [
            {
                title: "Oui",
                payload: "Oui",
            },
            {
                title: "Non",
                payload: "Non",
            },
        ],
    }, {
        default: true,
        handler: async (response, convo, bot) => {
            bot.say("It's dead, Jim.");
        },
    }, "confirmAide");

    controller.addDialog(dialog);
};
```

With this PR, it now just fails the type checking as it should.


designbyadrian:
How do we go around this problem before 4.6?

etiennellipse:
@designbyadrian It's mostly a syntactic issue when working in TypeScript.

Make sure to pass either a function or an array of handler objects (NOT a single handler object) and it works.

benbrown:
Thanks again!

PR #1821, opened 2019-10-03
Add support for Slack's new modals APIs and events

benbrown:
This PR adds support for [Slack's new modal APIS and events described here](https://api.slack.com/block-kit/surfaces/modals)

Changes include:

* Update dependency to latest version of @slack/web-api
* Add some new sample code to the testbot
* allow events without channels -- because new modal events do not include channel data. (This means bot.reply will not work with these events, amongst other caveats)

In addition, this will fix the behavior that caused all "block_actions" and "interactive_message" events to become "message" events.  These events will retain their native event type, but will still be passed into active dialogs (so a button click can be passed to a convo.ask)

### DIFF
--- a/packages/botbuilder-adapter-slack/package.json
+++ b/packages/botbuilder-adapter-slack/package.json
@@ -1,6 +1,6 @@
 {
   "name": "botbuilder-adapter-slack",
-  "version": "1.0.4",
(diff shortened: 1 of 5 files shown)


benbrown:
@geoff-watson yes

PR #1824, opened 2019-10-08
Add support for app home and modals

ryansydnor:
https://api.slack.com/block-kit/surfaces/modals

### DIFF
--- a/lib/SlackBot.js
+++ b/lib/SlackBot.js
@@ -345,7 +345,7 @@ function Slackbot(configuration) {
              * we do NOT want to respond to incoming webhooks or slash commands
              * as the response can be used by developers to actually deliver a reply
              */
-            if (!message.command && !message.trigger_word && !message.submission) {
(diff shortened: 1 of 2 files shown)


lib/SlackBot.js
```diff
@@ -345,7 +345,7 @@ function Slackbot(configuration) {
              * we do NOT want to respond to incoming webhooks or slash commands
              * as the response can be used by developers to actually deliver a reply
              */
-            if (!message.command && !message.trigger_word && !message.submission) {
+            if (!message.command && !message.trigger_word && !message.submission && message.type !== 'view_submission') {
```
ryansydnor: this means devs need to `bot.res.send('')` themselves because we don't have a `modal.Ok()` or `modal.error()` function.

ryansydnor:
Included in https://github.com/howdyai/botkit/pull/1842. Closing.

PR #1831, opened 2019-10-18
botbuilder-adapter-slack v1.0.5

benbrown:


packages/botbuilder-adapter-slack/src/slack_adapter.ts
```diff
@@ -524,12 +524,34 @@ export class SlackAdapter extends BotAdapter {
...
+                            activity.text = event.actions[0].selected_user;
+                            break
+                        case 'conversations_select':
+                            activity.text = event.actions[0].selected_conversation;
+                            break
+                        case 'channels_select':
+                            activity.text = event.actions[0].selected_channel;
+                            break
+                        case 'datepicker':
+                            activity.text = event.action[0].selected_date;
```
sixmen: I think, action -> actions

benbrown: ah you are correct!

PR #1833, opened 2019-10-22
End invalid Slack signature response.

kieran-bamforth:
When the Slack adapter receives an invalid signature header, it sets the status
code of the response to 401 but does not end the response.

This causes requests to hang that do not have a valid signature. This commit
therefore fixes the issue


ianfixes:
This helps with #1835

benbrown:
thanks! will be putting out another slack update soon.

PR #1836, opened 2019-10-24
Fix usePlugin parameter type declaration

etiennellipse:
This is a minor fix to the type declaration.

The current declaration is understood by TypeScript as `(botkit: Botkit) => (BotkitPlugin | BotkitPlugin)` while the intent is `((botkit: Botkit) => BotkitPlugin) | BotkitPlugin` (either a function producing a BotkitPlugin OR a BotkitPlugin.


benbrown:
Thanks for this!

PR #1850, opened 2019-11-07
Remove cms.ts from botkit package

etiennellipse:
This file is an outdated duplicate from the botkit-cms-plugin package, by all appearances.


benbrown:
thanks

PR #1854, opened 2019-11-12
Support ES6 syntax for botkit features

cooperka:
Now botkit can handle both CJS `module.exports` and ESM `export default` syntax.

Previously only this was supported:

```js
module.exports = function(controller) {
    controller.hears('sample', 'message', async (bot, message) => {
        await bot.reply(message, 'I heard a sample message.');
    });
}
```

Now you can also, optionally, use:

```js
export default (controller) => {
    // ...
}
```


cooperka:
[not stale]

benbrown:
I'll get this merged in to the next release.

PR #1855, opened 2019-11-12
Remove hbs dependency

Naktibalda:
Reasons:
* `hbs` view engine is registered, but not used anywhere.
* `handlebars` library had 2 security vulnerabilies recently, but `hbs` is slow to update,

We discussed it in Slack last week:
> benbrown 8:49 PM
It’s in the conversation system
> 
> naktibalda 10:03 AM
https://github.com/howdyai/botkit/blob/v4.6/packages/botkit/src/conversation.ts#L858 uses mustache, not handlebars
packages/botkit/src/conversation.ts:858

            outgoing.text = mustache.render(outgoing.text, { vars: vars });

> naktibalda 10:12 AM
The only package using hbs templates is docs, https://github.com/howdyai/botkit/tree/master/packages/docs
naktibalda 10:19 AM
`hbs` isn't used by botkit, but it could be used by third-party plugins so removing it is a BC-break and requires major version bump. Not worth it. (edited) 
benbrown 2:39 PM
Ah I think it would be safe
Plugins need to declare their own deps (edited)


Naktibalda:
[hbs: 4.1.0](https://www.npmjs.com/package/hbs) was released today,
it is necessary to either remove it from dependencies or update version constraint to allow upgrade to 4.1.0.

Naktibalda:
My previous comment was incorrect, there is no need to update version constraint to get a new version of `hbs`.
It would still be useful to remove this dependency.

benbrown:
I'm planning on building a new release as soon as #1894 is fixed.

PR #1867, opened 2019-11-24
Correct facebook channelData for fb_attachment from CMS

InnoraG:
Remove attachments from outgoing.channelData when generating fb_attachment especially on botkit CMS

No Harm code, just enable botkit CMS to correctly display fb_attachment carousel on emulator as well as messenger


benbrown:
the `fb_attachment` field that originates in the CMS should not be getting this far into the conversation system. This field should be removed during the loading process:

https://github.com/howdyai/botkit/blob/0ed3bc6b68adef015f9659c23a355329d635fd8e/packages/botkit-plugin-cms/src/cms.ts#L200

What are you trying to achieve here?

InnoraG:
Updated a missing additional commit required on botkit-plugin-cms to complete and clarify the pull.

Idea was to allow creation of dynamic fb_attachment (like messenger generic template carousel) from local or remote botkit CMS.

These change were required to make it work.

InnoraG:
Dear @benbrown 

Did you look at this PR by any chance ? It mainly add support for botkit-cms fb_attachment as function

Below line works only if fb_attachement is not a function, if it is a function below just fails
https://github.com/howdyai/botkit/blob/0ed3bc6b68adef015f9659c23a355329d635fd8e/packages/botkit-plugin-cms/src/cms.ts#L200

--------------------
1-So if it is a function it was not discarded , and similarly like other block type, additional check is required for the case fb_attachment is a function hence this change :
https://github.com/howdyai/botkit/blob/d089a66d353a4246f5ccc9683641e620f9a6b698/packages/botkit-plugin-cms/src/cms.ts#L176
suggested to introduce function exclusion just to avoid failure
```javascript
        if (line.fb_attachment && typeof (line.fb_attachment) !== 'function') {
```

--------------------

2-This change just certify fb_attachment is not included in slack or generic channelData.attachments
https://github.com/howdyai/botkit/blob/d089a66d353a4246f5ccc9683641e620f9a6b698/packages/botkit-plugin-cms/src/cms.ts#L170-L172
suggested to exclude line.fb_attachment when handling slack or generic
```javascript
        // move slack attachments
        if (line.attachments && !line.fb_attachment) {
            line.channelData.attachments = line.attachments;
```
-------------------
3-Accordinly when handling attachments function, now that fb_attachment function is supported
https://github.com/howdyai/botkit/blob/d089a66d353a4246f5ccc9683641e620f9a6b698/packages/botkit/src/conversation.ts#L823-L826
suggested to introduce required handling of fb_attachment function
```javascript
        if (typeof (line.attachments) === 'function') {
            // set both locations for attachments
            outgoing.attachments = outgoing.channelData.attachments = await line.attachments(line, vars);
            if (typeof (line.fb_attachment) === 'function') {
                    outgoing.channelData.attachment = await line.fb_attachment(line, vars);
                    delete outgoing.channelData.attachments;
            }
        }
```

--------------------
Not sure this PR is still valid, if required will make a new PR
Use case where fb_attachment as function is critical such as dynamic carousel, please double check, from my check it has no regression.
Users of botkit-cms are also welcome to test or comment.
Thank you

PR #1873, opened 2019-12-06
Support callbacks  in message template

qwertyuu:
As per this part of the documentation, dynamic generation of `quick_replies` can be achieved by passing a function as the value of specific properties of the `BotkitMessageTemplate` type. But, this type does not seem to support it in TypeScript (but everything goes well if I use a `//@ts-ignore` tag and the function is indeed called.)

So there you go, I edited the signature to fit the documentation.

If anyone wants to cover that part in tests, I think it would benefit everyone.

Thanks again for the nice botkit library 💃


benbrown:
ah good job!

maximebeaudoin:
@qwertyuu @benbrown we need to add the fields and the signature to the `attachment` and the `blocks` fields to the interface. This fields are currently missing and if we want to pass those to the `BotkitConversation` class it's failing because the signature don't allow those fields. According to the https://www.botkit.ai/docs/v4/conversations.html#dynamic-text-quick-replies-and-attachments documentation the should be supported.

Also, @qwertyuu can you please add some tests to the Dialog.tests.js file to covert this behavior ?

Thank you !

benbrown:
I will add those fields to the definition.

PR #1876, opened 2019-12-14
Modify onChange() example

mattcobb:
func arrow missing


benbrown:
this change should actually be made in the javascript file itself, as these reference docs are generated from code:

https://github.com/howdyai/botkit/blob/master/packages/botkit/src/conversation.ts#L486

mattcobb:
fix source comment instead

PR #1890, opened 2020-01-06
Bump to web-api 5.6.0

roeintense:
Exposes oauth.v2.access for new granular scoping.

I know this is tricky because the granular scoping still has a BETA tag on the Slack side, but bot installation fails because `getInstallLink()` and `validateOauthCode()` call against oauth.access, which throws `oauth_authorization_url_mismatch` when granular scoping is enabled on the Slack side.


benbrown:
addressed in 4.6.1

PR #1906, opened 2020-02-01
(Webex) Remove `files` key with empty string value

viveksyngh:
This commit removes sending `files` key with empty string value when
there are no files while sending a message. `files` key with empty string is causing API to fail
with below `BadRequest: "Unable to retrieve content."`. I have validated
that by sending request to webex api from webex developer console.

Signed-off-by: Vivek Singh <vivekkmr45@yahoo.in>

### DIFF
--- a/packages/botbuilder-adapter-webex/src/webex_adapter.ts
+++ b/packages/botbuilder-adapter-webex/src/webex_adapter.ts
@@ -277,9 +277,9 @@ export class WebexAdapter extends BotAdapter {
                     event: 'all',
                     secret: this.options.secret,
                     name: webhook_name
-                }).then(function() {
+                }).then(function () {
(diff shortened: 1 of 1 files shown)


viveksyngh:
I was trying webex adapter with botbuilder and I was able to get my bot app running and getting webhook callback from webex but it was not able to send the messages. It was failing with below error. 
```
BadRequest: Unable to retrieve content.
POST https://hydra-a.wbx2.com/v1/messages
WEBEX_TRACKING_ID: webex-js-sdk_b3d434d5-da91-48ac-bf92-99c05764affc_8
```

After some investigation,  I found that it was sending `files` key with empty even though there were no files. I tried the same payload in the webex developer console and there also it failed with same error. After I removed the `files` with empty value, it was able to send and receive messages.

benbrown:
Included in the next release -> #1929

PR #1919, opened 2020-02-24
add ouath v2 access endpoint

Jetersen:
for those wanted to use the new granular oauth scope


benbrown:
Merged and released! Thank you!

Issue #1924, opened 2020-03-10
Update conversations to parse channelData with mustache for CMS custom fields


me-cedric:
### What are you trying to achieve or the steps to reproduce?

I'm trying to pass custom fields containing conversation variables in a CMS conversation to trigger various api calls before sending messages.

### What was the result you received?

Custom fields are not parsed with mustache.

- Botkit version: 4.6.2
- Messaging Platform: Web/Slack
- Node version: v12.14.1
- Os: Windows 10 Enterprise 1803

PR #1925, opened 2020-03-10
Parse channelData with mustache in conversations

me-cedric:
#1924


benbrown:
This will be included in the next release -> #1929

PR #1927, opened 2020-03-10
add example response for oauth v2

sfny:
New slack apps will use oauth v2 by default, we can explain the discrepancy to help devs get up and running faster.


benbrown:
Excellent! Included in the next release -> #1929

Issue #1892, opened 2020-01-07
Add support for new Slack granular permission oauth endpoint


benbrown:
botbuilder-adapter-slack should be be patched to add support for using the new oauth/v2/authorize URL instead of generating a link to oauth/authorize

yingzhanguipath:
Installing new Slack apps is failing now: An API error occurred: oauth_authorization_url_mismatch

According to https://api.slack.com/methods/oauth.v2.access, 

> oauth_authorization_url_mismatch | The OAuth flow was initiated on an incorrect version of the authorization url. The flow must be initiated via /oauth/v2/authorize .

You'll probably need to use slack.oauth.v2.access instead of slack.oauth.access in validateOauthCode in slack_adapter.ts.

When do you think you can fix this issue?

benbrown:
For now I would directly use the methods available from the Slack SDK rather than using the built-in routes.

yingzhanguipath:
do you have an estimate on when you can release a fix?

benbrown:
Not currently. Since legacy apps will continue to need the current functions to remain the way they are.

You can generate your own redirect link, or use the Slack API to do it.

yingzhanguipath:
From Slack,
> Slack has improved how we handle permissions for apps, so you can now request only the specific scopes your app needs. We recommend updating your app’s scopes. Beginning **February 21**, all new app submissions to the App Directory must use granular permissions. Read our blog post.

All legacy apps will have to migrate soon. Once they migrate to new model, they need to "re-install", which will fail. Current botbuilder-adapter-slack is blocking the migration. You may want to reconsider your decision.

benbrown:
Hi @yingzhanguipath, I am not sure what is keeping you from generating your own redirect link and oauth validation method using the new URL and APIs.

All legacy apps that are *publicly listed in the app store* have to change, but the vast majority of Botkit apps are not public and will not be required to change in this timeline.  That is not to say that new methods won't be eventually introduced. You should consider opening a pull request!

mark-nimbix:
I am new to nodejs so I am sure there is a better way to do this but working on the yeoman generated slack app. I used the following in the bot.js. Hope it helps until there is a fix native. Oh and if you are using the yeoman generator you will need a newer version of botbuilder-adapter-slack. I used "botbuilder-adapter-slack": "latest" in package.json.

```
 async function validateOauthCodeNew(code){
    return new Promise(resolve => {
        var querystring = require('querystring');
[...]

PR #1928, opened 2020-03-10
adds oauth v2 support

sfny:
Changes to this file are based on an additional Slack adapter option of "oauthVersion". This would allow for oauth v2 support with minimal changes to the adapter. Updates to this file include 2 conditional statements to use new oauth v2 if options.oauthVersion == 'v2', and updates to example oauth response in the comments.

### DIFF
--- a/packages/botbuilder-adapter-slack/src/slack_adapter.ts
+++ b/packages/botbuilder-adapter-slack/src/slack_adapter.ts
@@ -244,7 +244,7 @@ export class SlackAdapter extends BotAdapter {
     public getInstallLink(): string {
         let redirect = ''
         if (this.options.clientId && this.options.scopes) {
-            if (this.options.oauthVersion == 'v2'|'V2'){
+            if (this.options.oauthVersion == 'v2'||'V2'){
                 redirect = 'https://slack.com/oauth/v2/authorize?client_id=' + this.options.clientId + '&scope=' + this.options.scopes.join(',');
             } 
             else {
--- a/packages/botbuilder-adapter-slack/src/slack_adapter.ts
+++ b/packages/botbuilder-adapter-slack/src/slack_adapter.ts
@@ -292,7 +292,7 @@ export class SlackAdapter extends BotAdapter {
             redirect_uri: this.options.redirectUri
         } 
         let results = {};
-        if (this.options.oauthVersion == 'v2'|'V2'){
+        if (this.options.oauthVersion == 'v2'||'V2'){
             results = await slack.oauth.v2.access(details)
         }
         else {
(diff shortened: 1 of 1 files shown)


sfny:
Attempting to solve #1892

packages/botbuilder-adapter-slack/src/slack_adapter.ts
```diff
@@ -280,12 +285,19 @@ export class SlackAdapter extends BotAdapter {
...
-        const results = await slack.oauth.access({
+        const details = {
             code: code,
             client_id: this.options.clientId,
             client_secret: this.options.clientSecret,
             redirect_uri: this.options.redirectUri
-        });
+        } 
+        let results = {};
+        if (this.options.oauthVersion == 'v2'|'V2'){
```
yingzhanguipath: I don't think this works.

this.options.oauthVersion === 'v2' || this.options.oauthVersion === 'V2'

yingzhanguipath: Do we really want to support both v1 and v2? All V1 will have to migrate to v2.

benbrown: YES, we need to support v1 until it is 100% disabled by Slack.

benbrown:
Included in the next release -> #1929

packages/botbuilder-adapter-slack/src/slack_adapter.ts
sfny:
```suggestion
        if (this.options.oauthVersion == 'v2'||'V2'){
```
this should suffice, good catch

packages/botbuilder-adapter-slack/src/slack_adapter.ts
```diff
@@ -242,9 +242,14 @@ export class SlackAdapter extends BotAdapter {
      * @returns A url pointing to the first step in Slack's oauth flow.
      */
     public getInstallLink(): string {
+        let redirect = ''
         if (this.options.clientId && this.options.scopes) {
-            let redirect = 'https://slack.com/oauth/authorize?client_id=' + this.options.clientId + '&scope=' + this.options.scopes.join(',');
-
+            if (this.options.oauthVersion == 'v2'|'V2'){
```
sfny:
```suggestion
            if (this.options.oauthVersion == 'v2'||'V2'){
```

benbrown: That is not a valid  conditional.

that is basically saying
`if ((this.options.authversion==v2) || ('V2'))`

it would need to be 2 separate conditions.

but i have done this in a different way so don't worry about it./

benbrown:
this was included in Botkit 4.8!

Issue #1920, opened 2020-03-03
Update handleTurn/spawn to accept optional alternate adapter


benbrown:
As it is, you can bind multiple adapters to the same Botkit controller.  [See this example](https://github.com/howdyai/botkit/blob/master/packages/testbot/multiadapter.js).

However, when the bot worker is spawned, it will always use the "default" controller.adapter that is passed in or created when `new Botkit()` is called.

We should introduce an optional parameter to the `handleTurn` call which accepts an adapter that will be used to spawn bots.  [This is the problematic code](https://github.com/howdyai/botkit/blob/master/packages/botkit/src/core.ts#L1075)

Issue #1911, opened 2020-02-07
Important slack event details missing from 'message' object.


BhuwanBhaskar:
Ref: https://github.com/howdyai/botkit/blob/a02291c7cef04b65ec5066bd8a8d287dd6c41115/packages/botbuilder-adapter-slack/src/slack_adapter.ts#L594

If slack events come as `event_callback` type then  'authed_users' list is missing from botkit 'message' object.
This is an important information which is missing as we use this list of authed users (for unfurl) to get user details from database.
I think whole payload of slack event should be attached in message object (like in previous botkit versions)

benbrown:
@BhuwanBhaskar ah, hrrm! That field is outside of the main payload.

We should at least copy it into channelData along with the main event.

benbrown:
This will be addressed in the next release -> #1929

PR #1929, opened 2020-03-11
Next Release

### LINKED ISSUE #1834: startConversationWithUser doesn't work with emulator
### Issue

According to [core docs](https://botkit.ai/docs/v4/reference/core.html#startConversationWithUser), `startConversationWithUser(reference: ConversationReference)` should work in the core without any adapter.

When I try to use it with the emulator, the async call to `client.conversations.createConversation(parameters)` inside `botworker.ts` seems to hang and the conversation is never started.

### Demo

I made a simple working demo app [here](https://github.com/cooperka/botkit-issue-demos/commits/2019-startConvoWithUser) that you can clone to see the issue.

1. Start the emulator
2. Say "ref" to save a conversation reference to be used later (see console log output)
3. The bot should automatically start a new conversation using the ref, but instead it hangs

### Debugging info

Apparently after about 5 minutes, the socket hangs up with this dump:

```console
Error: socket hang up
    at new RestError (botkit-issue-demos/node_modules/@azure/ms-rest-js/dist/msRest.node.js:1397:28)
```

```js
{
  code: 'REQUEST_SEND_ERROR',
  statusCode: undefined,
  request:
   WebResource {
     url: 'http://localhost:52343/v3/conversations',
     method: 'POST',
     body:
      '{"isGroup":false,"bot":{"id":"342e1f30-f4fa-11e9-aad8-0d5bc2ce5f40","name":"Bot","role":"bot"},"members":[{"id":"a97ee945-eace-4201-a682-a45e2ed2af22","name":"User","role":"user"}],"activity":null,"channelData":null}',
  ...
}
```

It's unclear from the async stack trace where this request originates.

### Context

- Botkit version: 4.6.1
- Messaging Platform: Emulator
- Node version: 10.15.3
- OS: Mac 10.14

### LINKED ISSUE #1878: bot.beginDialog causes unexpected message if dialog already open
### Issue

Using `bot.beginDialog` or `bot.replaceDialog` when there's already a dialog open (e.g. to switch to the new dialog based on user input) causes the bot to automatically receive an unexpected `{}` message.

This message breaks questions, because the empty message is interpreted by botkit as an answer to the question, and the user doesn't have a chance to give input.

### Demo

I made a simple working demo app [here](https://github.com/cooperka/botkit-issue-demos/commits/2019-dialogJumping) that you can clone to see the issue.

1. Start the emulator
1. Say `hi` to trigger ONBOARDING_DIALOG
1. Say anything for your name
1. Say `yes` to confirm and trigger ONBOARDING_DIALOG_2
1. onboarding2 asks for your name, and immediately receives `{}` as a response without your input. It's impossible to reply with your name.

### Workaround

See [this commit](https://github.com/cooperka/botkit-issue-demos/commit/16abd7a76f676f7c8f9a7c4db72c2de21de55b14) for a successful workaround using `dialog.addGotoDialog` instead of `bot.beginDialog`.

It's great that this works, but it isn't possible to use this in all cases (e.g. [this issue](https://github.com/howdyai/botkit/issues/1874#issuecomment-567296864)).

### Context

- Botkit version: 4.6.1
- Messaging Platform: Emulator
- Node version: 10.15.3
- OS: Mac 10.14

### LINKED ISSUE #1932: Calling beginDialog from inside a `ask()` handler causes a message to be skipped
Due to the way the dialog stack works, calling a `bot.beginDialog()` inside an `ask()` handler will cause the next step of the dialog in which the `ask()` is called to be skipped.

This is because the `ask()` handler actually fires during the first phase of processing that next step, and when the beginDialog occurs, the process is short circuited. There is currently no way to resume at that step after the child dialog has completed.

One possible solution is to split the message into 2 actual steps, one solely for processing the response to the question and a second to send the following message.  This can theoretically be achieved inside the .ask function.

benbrown:
* Full message passed to dialog `ask` handlers #1801 #1973 
* Multi-adapter support improved #1920 
* Support for slack v2 oauth #1892 #1928 
* Fix for Webex sending with empty files array #1906 
* Parse channelData field as mustache templates #1925 
* move authed_users into message object #1911 
* fix #1834 
* fix #1878 
* fix #1932


GaryTowers:
Excited for this! 🙌I'm updating my bot from botkit 0.4 to latest and also updating scopes using V2 OAuth and this release a critical peace of it

PR #1944, opened 2020-03-23
Removes enforcement of "as_user" field. Fixes #1943

### LINKED ISSUE #1943: "as_user" has been deprecated by Slack but Botkit still uses it
I'm trying to send a message from my bot with a custom avatar using the `icon_url` property

````javascript
await bot.reply({
   text: "Hey this is me",
   icon_url: "http://lorempixel.com/48/48/" //<-- Botkit reads this and forces the use of a deprecated 'as_user' field
});
````

I received an error message form the Slack API that it is tied with the `as_user` when using `icon_url` in the reply object.
```
code:"slack_webapi_platform_error"
data:Object {ok: false, error: "invalid_arguments", deprecated_argument: "as_user", …}
message:"An API error occurred: invalid_arguments"
stack:"Error: An API error occurred: invalid_arguments
```

### What did you expect?
I expected that the `icon_url` field could be used without the `as_user` field, because the `icon_url` is now dependent over Slack's [new granular bot permissions](https://api.slack.com/authentication/basics#public).

However the Botkit Slack Adapter is still assuming the field `as_user` is required to be sent as `false` when `icon_url` is used on replies [(slack_adapter.ts file)](https://github.com/howdyai/botkit/blob/7bcb10033c1c9cfa319e398b0f012aafcb207275/packages/botbuilder-adapter-slack/src/slack_adapter.ts#L343) as you would do with a previous generation **classic bot**.

Slack mentions that the `as_user` field [(may not be used on new generation bots)](https://api.slack.com/methods/chat.postMessage#arg_as_user)

My recommendation to fix this would be to remove the [(enforcement of the IF statement)](https://github.com/howdyai/botkit/blob/7bcb10033c1c9cfa319e398b0f012aafcb207275/packages/botbuilder-adapter-slack/src/slack_adapter.ts#L343) setting `as_user` to `false` when setting the field `icon_url` and leave it to the developer to manually include the `as_user` on their reply object if ever needed.

### Context:

- Botkit version: 4.8.1
- Messaging Platform: Slack
- Node version: 13.7.0
- Os: macOS Mojave

GaryTowers:
Fixes issue #1943 

By removing this IF statement: 

**New generation bots** will be able to make use of `icon_url`, `icon_emoji` and `username` fields on the `reply` object without having Botkit forcing a deprecated field `as_user` set to false.

For **classic bots**, developers would still be able to add the `as_user` field manually as stated by the [Slack documentation](https://api.slack.com/methods/chat.postMessage#arg_as_user)

_(I'm not sure if I should make a PR to master or other branch, please advice)_

### DIFF
--- a/packages/botbuilder-adapter-slack/src/slack_adapter.ts
+++ b/packages/botbuilder-adapter-slack/src/slack_adapter.ts
@@ -340,6 +340,14 @@ export class SlackAdapter extends BotAdapter {
             message.user = activity.recipient.id;
         }
 
+        if (message.icon_url || message.icon_emoji || message.username) {
+            message.as_user = false;
+        }
+
+        // as_user flag is deprecated on v2
+        if (message.as_user === false && this.options.oauthVersion === 'v2') {
+            delete message.as_user;
+        }
 
         return message;
     }
--- a/packages/botbuilder-adapter-slack/src/slack_adapter.ts
+++ b/packages/botbuilder-adapter-slack/src/slack_adapter.ts
@@ -340,9 +340,6 @@ export class SlackAdapter extends BotAdapter {
             message.user = activity.recipient.id;
         }
 
-        if (message.icon_url || message.icon_emoji || message.username) {
-            message.as_user = false;
-        }
 
         return message;
     }


benbrown:
This would break existing bots using this feature.

Please add a check for the new `oauth_version` configuration option - only exclude field if it is set to `v2`

GaryTowers:
Updated to fix issue #1943 based on `oauth_version` set to `v2`

The v2 check before removing the `as_user` field makes this backwards compatible with v1

Thanks for the observation @benbrown

Issue #1945, opened 2020-03-23
controller.spawn() throws UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'context' of undefined


NxP4Code:
Steps to reproduce
Run the following sample code (https://hazel-jazzy-peach.glitch.me/):
```

//  __   __  ___        ___
// |__) /  \  |  |__/ |  |  
// |__) \__/  |  |  \ |  |  

// This is the main file for the starter-webex bot.

// Import Botkit's core features
const { Botkit } = require('botkit');
const { BotkitCMSHelper } = require('botkit-plugin-cms');

// Import a platform-specific adapter for webex.

const { WebexAdapter } = require('botbuilder-adapter-webex');

const { MongoDbStorage } = require('botbuilder-storage-mongodb');
const common = require('./common');


// Load process.env values from .env file
require('dotenv').config();

let storage = null;
if (process.env.MONGO_URI) {
    storage = new MongoDbStorage({
        url : process.env.MONGO_URI,
    });
}


const adapter = new WebexAdapter({
    // REMOVE THIS OPTION AFTER YOU HAVE CONFIGURED YOUR APP!
    enable_incomplete: true,
  
    access_token: process.env.access_token,
    public_address: process.env.public_address,
    secret: process.env.secret
})    


const controller = new Botkit({
    webhook_uri: '/api/messages',

    adapter: adapter,

    storage
});

controller.spawn().then(bot => {
    common.bot = bot;
     bot.api.messages.create({
       text: "beep, boop!",
       toPersonEmail: 'ABC.DEF@here.com'

     });
  });

if (process.env.cms_uri) {
    controller.usePlugin(new BotkitCMSHelper({
        cms_uri: process.env.cms_uri,
        token: process.env.cms_token,
    }));
}

controller.usePlugin({
  name: 'webserver logger',
  init: function(botkit) {
    console.log('binding new express middleware');
    botkit.webserver.use(function(req, res, next) {
      console.log('FIRE MIDDLEWARE');
      console.log('> ', req.url, JSON.stringify(req.body, null, 2));
      next();
    });
  }
});

// Once the bot has booted up its internal services, you can use them to do stuff.
controller.ready(() => {

    // load traditional developer-created local custom feature modules
    controller.loadModules(__dirname + '/features');

    /* catch-all that uses the CMS to trigger dialogs */
    if (controller.plugins.cms) {
        controller.on('message,direct_message', async (bot, message) => {
            let results = false;
[...]

PR #1948, opened 2020-03-26
Fixed bug-1945

NxP4Code:
I have made the changes in core.ts file to fix [Issue-1945](https://github.com/howdyai/botkit/issues/1945)


.gitignore
```diff
@@ -20,3 +20,4 @@ coverage
 
 # temp test files
 tests/**/temp/
+package-lock.json
```
benbrown: don't ignore package-lock

packages/botkit/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "botkit",
-  "version": "4.8.1",
+  "version": "4.8.2",
```
benbrown: i'll handle this don't worry about changing version number

NxP4Code: I reverted the version.

PR #1960, opened 2020-04-22
Fix dynamic text field type definition

etiennellipse:
See [BotkitConversation.makeOutgoing](https://github.com/howdyai/botkit/blob/master/packages/botkit/src/conversation.ts#L795-L802) method. 

The `text` variable must be a string, as it is fed into the mustache renderer [here](https://github.com/howdyai/botkit/blob/master/packages/botkit/src/conversation.ts#L867). So the template function must always return a string, or the function will crash later on.

Therefore, the type definition must reflect that. I am getting TypeScript error when returning a string, if I can it to an array, botkit breaks. 😞

### DIFF
--- a/packages/botkit/src/conversation.ts
+++ b/packages/botkit/src/conversation.ts
@@ -36,7 +36,7 @@ interface BotkitConvoTrigger {
  * Template for definiting a BotkitConversation template
  */
 interface BotkitMessageTemplate {
-    text: ((template: any, vars: any) => string[]) | string[];
+    text: ((template: any, vars: any) => string) | string[];
     action?: string;
     execute?: {
         script: string;


benbrown:
If specified as an array, the function will pick one at random.  What breaks if you do this?

etiennellipse:
@benbrown It works as expected, no problem with that. 

The issue is when passing a function with a template, the function must return only one string (and not an array). It is how botkit works, but the type definition does not reflect that correctly. So when using Botkit with a TypeScript project, the type checker screams.

benbrown:
oh i see, only when applied to the FUNCTIONS.  got it! Merging!

etiennellipse:
Yeah, sorry it was not clear in my PR description now that I read it again.

PR #1967, opened 2020-05-08
Missing underscores

somelinuxguy:
/lib/slack_adapter.js is looking for CLIENT_ID and CLIENT_SECRET


benbrown:
thanks! this change will be part of the next release.

PR #1979, opened 2020-06-10
Fix channelData replace in conversations

ashitikov:
https://github.com/howdyai/botkit/pull/1925 introduced new bug:
When I'm trying to use channelData: { ephemeral: true } with dynamic blocks I get an empty ephemeral message in slack because channelData completely replaces instead of correct copying keys and values with template rendering.

### DIFF
--- a/packages/botbuilder-adapter-slack/CHANGELOG.md
+++ b/packages/botbuilder-adapter-slack/CHANGELOG.md
@@ -1,5 +1,9 @@
 # botbuilder-adapter-slack changelog
 
+# 1.0.12
+
+* Adjust startPrivateConversation to use converations.open instead of im.open
+
 # 1.0.11
 
 * Update dependencies to Botkit 4.9, Bot Framework 4.9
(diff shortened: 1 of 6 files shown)


benbrown:
can you give me more information on how to reproduce this issue?

ashitikov:
> can you give me more information on how to reproduce this issue?

Hi! If you have a dialog
`const dialog = new BotkitConversation('ROOM_DETACH', controller);`
Then you can add a question to this conversation:
`dialog.addQuestion({
    text: 'How are you?',
    channelData: {
        ephemeral: true
    },
    blocks: async (template, vars) => {
        return [...some blocks...]
    }
});
`
As a developer I'm waiting that this question 'How are you?' will be ephemeral (only visible to that user) with blocks, but in fact this message does not contain any blocks, but becomes ephemeral.

Commenting out channelData { ephemeral: true } makes message not ephemeral, but blocks are visible as it should.

ashitikov:
> ah i see, thank you for the info.

Any progress on that? :)

PR #1981, opened 2020-06-18
Add Block Kit library links to README

raycharius:
As discussed in Bot Developer Hangout, adding links to two libraries that make it easier to compose Slack UIs.

Of the ones available, these were the two that are actively maintained and are documented well enough. 

There is another – [slack-block-msg-kit](https://github.com/IyiKuyoro/slack-block-msg-kit) – but just felt a bit verbose to me. Also wasn't a fan of it validating Slack business logic, without configuration available, to bypass the validation. Could create limitations and issues when Slack updates their own validation rules.


raycharius:
Thanks, Ben!

PR #1983, opened 2020-06-19
Fix unhandled promise rejections when a trigger errors

abrenneke:
When a trigger, such as the callback to `controller.hears(...)` errors (the async callback rejects), this promise is not handled correctly, and results in an unhandled promise rejection.

This is because the `run` function for [`ware`](https://github.com/segmentio/ware) does not support promises (the middleware themselves do, but not the callback). 

I don't really agree with the decision to use a non-promise-supporting library like `ware` in this very promise-based framework, but we can at least promisify those calls.

This ends up simplifying a lot of the code that uses the middleware, as Promise constructors no longer need to be used.

With this fix, the promises bubble up all the way to the `handleTurn` function, and can be handled at the web layer.

Maybe a `controller.on('error', () => {})` event would make sense as well. Or an `error` middleware?

### DIFF
--- a/packages/botkit/src/botworker.ts
+++ b/packages/botkit/src/botworker.ts
@@ -97,16 +97,9 @@ export class BotWorker {
      * @returns Return value will contain the results of the send action, typically `{id: <id of message>}`
      */
     public async say(message: Partial<BotkitMessage> | string): Promise<any> {
-        return new Promise((resolve, reject) => {
-            const activity = this.ensureMessageFormat(message);
-
-            this._controller.middleware.send.run(this, activity, async (err, bot, activity) => {
-                if (err) {
-                    return reject(err);
-                }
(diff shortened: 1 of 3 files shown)


abrenneke:
Now that I think about this a bit more, if a middleware doesn't call `next()` (i.e. the middleware ends the request) then there will be a promise that's never resolved, so memory/handler leak. I think the original code had this problem too because it was using a Promise constructor that ended up never resolving. Maybe that should be addressed separately.

alex-ubitec:
I've encountered similar problems as well.

alex-ubitec:
That PR doesn't fix the problem. 

Please take a look at the promises that this PR tries to change, because they are simply written incorrectly. The suggestion to use promisify is just to prevent those errors from propping up again, but you could just wrap the code inside the `new Promise` calls with a try-catch as well.

PR #1986, opened 2020-07-01
Result is not necessarily defined here

alex-ubitec:
You can get an undefined result at this point, since `runStep` can be called with an empty result. I don't believe it is assumed that an action handler is called always after the previous step generated a result

Until this is fixed, what would be an alternative?


benbrown:
thanks for this fix! it will be released in 4.10.

PR #2004, opened 2020-07-22
Support updating a Slack message with blocks

keslerm:
Minor change to include copying the `blocks` property to allow updating a blocks based message.

I didn't include any tests with this as it's a somewhat trivial change but can if need be.


keslerm:
Actually just realized that specifying `blocks` within `channelData` gets the right info into Botbuilder - It's just non-obvious.

Found from https://github.com/microsoft/botframework-sdk/issues/5434

Example:

```
    await bot.updateMessage({
      channelData: {
        blocks: [
          {
            "type": "section",
            "text": {
              "type": "mrkdwn",
              "text": `:loading: Deploying ${message.matches[1]} to ${params.get("env")}...`
            }
          }
        ]
      }
    }, ...sent);
```

PR #2008, opened 2020-07-30
Included next call middleware. Fixes #2007

### LINKED ISSUE #2007: [slack] Web middlewares to /api/messages get ignored
Custom webserver middlewares for the `/api/messages` endpoint get ignored 

Yes. As stated in the documentation, the **webserver** is based off an expressjs app, meaning middlewares can be added to it through `.use(...)` (get, post, delete or update as well)

1. Create a middleware for the `/api/messages` endpoint
````javascript
controller.webserver.use('/api/messages', (req, res, next) => {
   // My custom middleware code
   next();
})
````
2. Trigger a post call to the endpoint (using slack or postman)
3. The bot receives the event on the endpoint, but the custom middleware never gets called

The middleware created for the endpoint never gets called.

This is most likely to happen since the [botkit adapter contains an internal middleware that never calls](https://github.com/howdyai/botkit/blob/848712b5d11ec4b8da3a5a0bf4e4f4aa4742c43f/packages/botkit/src/core.ts#L690) `next()` to pass control over to other custom created middlewares.

### What did you expect?

That the custom middleware created for `/api/messages` gets called

### Possible solution

Include a `next()` call in[ the definition of Botkit's ](https://github.com/howdyai/botkit/blob/848712b5d11ec4b8da3a5a0bf4e4f4aa4742c43f/packages/botkit/src/core.ts#L690)`/api/messages` endpoint middleware. This would allow other custom middlewares to be called while remaining consistent with expressjs middleware functionality.

### Context:

- Botkit version: 4.8.1
- Messaging Platform: Slack
- Node version: v14.0.0
- Os: macOS Mojave 10.14.6

GaryTowers:
Fixes issue #2007 allowing adding middlewares to `/api/messages` endpoint following the behavior of a expressjs middlewares

### DIFF
--- a/packages/botkit/src/core.ts
+++ b/packages/botkit/src/core.ts
@@ -687,7 +687,7 @@ export class Botkit {
      */
     private configureWebhookEndpoint(): void {
         if (this.webserver) {
-            this.webserver.post(this._config.webhook_uri, (req, res) => {
+            this.webserver.post(this._config.webhook_uri, (req, res, next) => {
                 // Allow the Botbuilder middleware to fire.
                 // this middleware is responsible for turning the incoming payload into a BotBuilder Activity
                 // which we can then use to turn into a BotkitMessage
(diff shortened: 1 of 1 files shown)


benbrown:
Instead of this, you should use the webserver_middlewares configuration parameter that lets you bind additional middlewares _before_ the endpoint handlers are added.

PR #2022, opened 2020-08-21
Version 4.10 WIP

### LINKED ISSUE #1961: Expose helpers for MS Teams
We should provide a BotWorker subclass that adds some teams helpers like:

* updateActivity
* access to new TeamsInfo methods from Bot Framework
* others?

### LINKED ISSUE #1984: Broken statement in core.ts when selecting the adapter in the spawn() function of version 4.9.0
While trying to pinpoint an issue with using a custom adapter and worker within unit tests, I came upon a change that happened in #1938 on core.ts [right here](https://github.com/howdyai/botkit/pull/1938/files#diff-55821637a45d41c6f81a38d283fc253aR1080). It is in the spawn function of the core.ts file, when assigning the `adapter` variable.

[Latest version, line 1093](https://github.com/howdyai/botkit/blob/main/packages/botkit/src/core.ts#L1093)

This condition seems faulty:
````
const adapter = custom_adapter || (config && config.context && config.context.adapter) ? config.context.adapter : this.adapter;
````

It will never return the `custom_adapter`, if one was passed. Since the statement is now a ternary, the custom_adapter will not be returned. 

As in this example:
```
true || (false) ? "1" : "2"
```
The returned value is `"1"`, not `true`.

To get something similar to the previous behaviour, the line should look something like:
```
const adapter = custom_adapter || ((config && config.context && config.context.adapter) ? config.context.adapter : this.adapter);
```

I cannot test this since I have no idea how to pass a custom_adapter to the spawn function! I did not find any test code using it.

### Context:

- Botkit version: 4.9.0

### LINKED ISSUE #2003: [botbuilder-adapter-web] TypeError: cb is not a function

Yes (botbuilder-adapter-web)

Keeping a WebSocket running.

````
// Wrap code in markdown source tags
````

The above error, with the following stack trace:

````
TypeError: cb is not a function
1
at sendAfterClose (/app/packages/api/node_modules/ws/lib/websocket.js line 733 col 4)
cb(err);
2
at WebSocket.ping (/app/packages/api/node_modules/ws/lib/websocket.js line 273 col 6)
sendAfterClose(this, data, cb);
3
at each (/app/packages/api/node_modules/botbuilder-adapter-web/lib/web_adapter.js line 144 col 19)
ws.ping('', false, true);
4
at Set.forEach (<anonymous> line (unknown))
5
at Timeout._onTimeout (/app/packages/api/node_modules/botbuilder-adapter-web/lib/web_adapter.js line 139 col 29)
this.wss.clients.forEach(function each(ws) {
6
at listOnTimeout (internal/timers.js line 549 col 16)
7
at processTimers (internal/timers.js line 492 col 6)
````

https://github.com/howdyai/botkit/blob/848712b5d11ec4b8da3a5a0bf4e4f4aa4742c43f/packages/botbuilder-adapter-web/src/web_adapter.ts#L160
https://github.com/websockets/ws/blob/da42ea17451f11eed54adb54d3beeedbb1c2aa70/lib/websocket.js#L270
https://github.com/websockets/ws/blob/da42ea17451f11eed54adb54d3beeedbb1c2aa70/lib/websocket.js#L730

### What did you expect?

No Error

### Context:

- Botkit version:
- Messaging Platform: Web
- Node version: v12.18.2
- Os: Ubuntu 18.04.4 LTS
- Any other relevant information:

### Analysis

IMO the call

````
ws.ping('', false, true);
````

is faulty since the third argument should be a function.

benbrown:
fixes #1984 
fixes #1961 
fixes #2003

### DIFF
--- a/packages/botbuilder-adapter-slack/src/slack_adapter.ts
+++ b/packages/botbuilder-adapter-slack/src/slack_adapter.ts
@@ -172,7 +172,7 @@ export class SlackAdapter extends BotAdapter {
 
         this.middlewares = {
             spawn: [
-                async (bot, next) => {
+                async (bot, next): Promise<void> => {
                     // make the Slack API available to all bot instances.
                     bot.api = await this.getAPI(bot.getConfig('activity')).catch((err) => {
                         debug('An error occurred while trying to get API creds for team', err);
--- a/packages/botbuilder-adapter-web/src/web_adapter.ts
+++ b/packages/botbuilder-adapter-web/src/web_adapter.ts
@@ -157,7 +157,9 @@ export class WebAdapter extends BotAdapter {
                     return ws.terminate();
                 }
                 ws.isAlive = false;
-                ws.ping('', false, () =>{});
+                ws.ping('', false, () => {
+                    // noop
+                });
             });
         }, 30000);
     }
--- a/packages/botbuilder-adapter-webex/src/botworker.ts
+++ b/packages/botbuilder-adapter-webex/src/botworker.ts
@@ -82,7 +82,7 @@ export class WebexBotWorker extends BotWorker {
             channelId: 'webex'
         });
     }
-    
+
     /**
      * Switch a bot's context into a specific thread within a room.
      * After calling this method, messages sent with `bot.say` and any dialogs started with `bot.beginDialog` will occur in this new context.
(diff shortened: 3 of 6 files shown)


Danjuro (commented):
This looks great exposing the TeamsInfo api, however is there a reason why you don’t update bot.updateMessage to work for teams this way such as https://botkit.ai/docs/v4/reference/slack.html#updatemessage

(I am going to have to double check this method now in teams as I didn’t think it worked before)

benbrown:
@Danjuro good question! I don't have a great answer. My initially reasoning is that the Bot Framework adapter can be used for other platforms in addition to Teams, and they may not support those methods. 

That said, I'll see what I can do!

PR #2040, opened 2020-09-18
adds Slack WebClientOptions

iamtofr:
I just wanted to use a proxy on the Slack WebClient. But not passing down all options to the WebClient makes it impossible to configure the Client.


benbrown (changes requested):
There are a few more places in the code where the WebClient is created.  Can you also apply this change to those?

iamtofr:
Sure i will take a look

iamtofr (commented):
adds configuration object to missing instantiating WebClients

PR #2046, opened 2020-11-25
Legacy support for Slack oauth v2

richardqhill:
https://api.slack.com/authentication/oauth-v2


ryansydnor (approved):
LGTM. @benbrown where is the documentation for the legacy branch so we can include this configuration option?

ryansydnor:
Here's where we can update the docs: https://github.com/howdyai/botkit-docs/blob/master/docs/index.md

Issue #2217, opened 2022-05-20
Uncatchable throws on botbuilder-adapter-webex registerWebhookSubscription Functions


punithk:
We were using botbuilder-adapter-webex and I see there was an issue where we change or added wrong tokens from Webex the application crashes. There was no way to catch the error as the throw error was after asynchronous operation and that could not be caught by try catch.

What we tried was:
````javascript
const adapter = new WebexAdapter({
     access_token: process.env.ACCESS_TOKEN, // access token from https://developer.webex.com
     public_address: process.env.PUBLIC_ADDRESS,  // public url of this app https://myapp.com/
     secret: process.env.SECRET // webhook validation secret - you can define this yourself
});

adapter.registerWebhookSubscription('/api/messages'); // <- Code crashes here and try catch does not help when tokens are wrong
adapter.getIdentity();

````

What we tried was:
````javascript
const adapter = new WebexAdapter({
     access_token: process.env.ACCESS_TOKEN, // access token from https://developer.webex.com
     public_address: process.env.PUBLIC_ADDRESS,  // public url of this app https://myapp.com/
     secret: process.env.SECRET // webhook validation secret - you can define this yourself
});
try{
    adapter.registerWebhookSubscription('/api/messages'); // <- This does not help
    adapter.getIdentity();
}
catch(err){
    console.error(err);
}

````

### What was the result you received?
Code crashed even with try catch.

### What did you expect?
The error should be catchable on try catch or .then().catch(); so that it can be managed. If tokens expire or are changed it would simply crash the server which should not be the behavior instead it should be handled to gracefully let user decide.

### Probable reason its happening:
Both ````registerWebhookSubscription```` and ````registerAdaptiveCardWebhookSubscription```` are not returning promise and throwing error asynchronously, wrapping it with a promise should fix it.

### Context:

- Botkit version: any version
- Messaging Platform: Any
- Node version: Any
- Os: Any
- Any other relevant information:
Try to wrap and provide solution as a pull request

Dayavats:
facing similar issue while using this package  in case of wrong token

pgoldweic:
[...]

PR #2218, opened 2022-05-20
Add Promise to botbuilder-adapter-webex both registerWebhookSubscription functions

punithk:
This is a fix for [issue 2217](https://github.com/howdyai/botkit/issues/2217).

The solution wraps the function with promise such that rejects are present and any thrown error can not be easily caught and handled by the user and should not affect existing bots.

Now developers can handle wrong access tokens this way without crashing during webhook subscription:

```javascript
const adapter = new WebexAdapter({
     access_token: process.env.ACCESS_TOKEN, // access token from https://developer.webex.com
     public_address: process.env.PUBLIC_ADDRESS,  // public url of this app https://myapp.com/
     secret: process.env.SECRET // webhook validation secret - you can define this yourself
});

adapter.registerWebhookSubscription('/api/messages')
     .then()
     .catch(err => {    // <- this would fix the issue
            // Handle the error as per needs
     });
```

The changes should fix the issue and be backward compatible too.

### DIFF
--- a/packages/botbuilder-adapter-webex/src/webex_adapter.ts
+++ b/packages/botbuilder-adapter-webex/src/webex_adapter.ts
@@ -253,52 +253,56 @@ export class WebexAdapter extends BotAdapter {
      * Register a webhook subscription with Webex Teams to start receiving message events.
      * @param webhook_path the path of the webhook endpoint like `/api/messages`
      */
-    public registerWebhookSubscription(webhook_path): void {
-        const webhook_name = this.options.webhook_name || 'Botkit Firehose';
+    public registerWebhookSubscription(webhook_path): Promise<any> {
+        return new Promise<void>((resolve, reject) => {
+            const webhook_name = this.options.webhook_name || 'Botkit Firehose';
 
-        this._api.webhooks.list().then((list) => {
-            let hook_id = null;
+            this._api.webhooks.list().then((list) => {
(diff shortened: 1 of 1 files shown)


Dayavats:
Promisifying should resolve the issue of application breaking on wrong token case

Akanksha-270392:
Thanks for the suggestion. Looks like it will solve the issue.

VaniKaushik-2511:
The above suggestion is appreciated. It look like it will be the solution to the issue.

Maleeha456:
The solution mentioned above seems to resolve the issue.

Sayak-Bhattacharjee:
It seems like this solution will solve the issue. Appreciated.

Dhananjay220398:
The solution mentioned above seems to resolve the issue.Thanks for the suggestion.

Adishjain58:
Thanks for the suggestion. Looks like it will solve the issue.

Akash2695:
Seems like this solution will resolve the issue, Thanks for the suggestion.

rashmi0403:
Thanks for the suggestion. Looks like it will solve the issue.

Sakshi21197:
I think @punithkrishnamurthy is doing right thing can we get this merged?

Deekshashandilya:
facing similar issue while using this package. Please suggest some solution.

Issue #2221, opened 2022-06-10
Webex Teams Adapter Doesn't Support direct_mention Event Type


bbrougher-cerium:
### What was the result you received?

When trying to have my bot respond to a ```direct_mention``` event, nothing happened. The event came across into the express server and my controller event was never triggered.

### What did you expect?

I expected my controller event handler (snippet below) to run, but nothing happened.

``` typescript
import { Botkit } from "botkit";
import { WebexAdapter } from "botbuilder-adapter-webex";
import { Express } from "express";

const adapter = new WebexAdapter({
  enable_incomplete: true,
  access_token: process.env.ACCESS_TOKEN,
  public_address: process.env.PUBLIC_ADDRESS,
});

const controller = new Botkit({
  webhook_uri: "/api/messages",
  adapter,
});

controller.on("direct_mention",
  async (bot, message): Promise<void> => {
    console.log("Received message: ");
    console.info(message);
    await bot.reply(message, "Hello!");
  }
);

controller.ready(() => {
  console.log("Bot loaded.");
});

const app: Express = controller.webserver;

app.get("/", (req, res) => {
  res.send(`This app is running. Botkit version: ${controller.version}`);
});
```

### Context:

- Botkit version: 4.15.0
- Messaging Platform: Webex Teams
- Node version: v18.0.0
- OS: Windows
- Any other relevant information:

I have a PR with some small code changes that seem to be working for my use case that I will submit shortly.

PR #2222, opened 2022-06-10
Add direct_mention support to Webex Teams adapter

bbrougher-cerium:
I submitted an issue about this here: #2221 

This is the code I was able to use to get my bot working and responding to the ```direct_mention``` event type. I'd be happy to expand on this if there are edge cases that aren't covered by this or if there are tweaks that need to be made.


NxP4Code (approved):
Changes looks good

NxP4Code (approved):

PR #2239, opened 2023-02-14
Update dependencies and replace `request` package

### LINKED ISSUE #2238: Remove deprecated request package

In several of the `botkit` adapters, there are imports for the `request` package - [it's been deprecated for a few years now](https://github.com/request/request#deprecated).

The `request` library should be replaced with a package that is maintained. For Node.js 18+, there's a native `fetch` that can be used, but for other use cases [the `node-fetch` library](https://github.com/node-fetch/node-fetch) is great (or `cross-fetch` for isomorphic uses). Plus, switching to it makes it easier in the future to switch to the native Node / browser `fetch` API.

strmer15:
This fixes https://github.com/howdyai/botkit/issues/2238

* Updates a lot of dependencies, since I couldn't actually install or build anything until I did a lot of these
  - Latest `eslint` and plugins, so that there weren't peer dependency errors
    - Updated the config files, notably switched `@typescript-eslint/ban-ts-ignore` to `@typescript-eslint/ban-ts-comment`
    - `eslint-plugin-node` has been replaced by `eslint-plugin-n`
  - Latest `typescript` and `tsdoc`, `tsdoc` 0.22 was giving peer dependency errors about old `typescript`
    - Fixed the `build-docs` command, seems like it may have been broken for a while - incorrect options, missing tsconfig
    - Removed `typedoc-plugin-external-module-name` since it's no longer maintained and doesn't work with newer `tsdoc` versions
  - Regenerated `package-lock.json` files, some of them were deleted since they're no longer used
  - Updated to latest `lerna` and updated the `readme`
 
* Replaced `request` with `cross-fetch` in a few different places
  - https://github.com/lquixada/cross-fetch
  - The trickiest part of this is handling the errors and responses - in each case, I'm checking `response.ok` (for a 2xx response status) and then throwing an `Error` if it doesn't match. Then, I'm handling the response as plain text or as JSON if it parses correctly.


strmer15:
@microsoft-github-policy-service agree company="The Farmer's Dog"

PR #2243, opened 2023-06-21
Enhanced FB Workplace Compatibility and added Workplace Group Thread support

marcopolo39:
- Fixed an issue where bot would receive random requests from workplace which would cause the bot to crash. 
- Added support for workplace threads by extracting thread from incoming request and replacing user id to thread_id in outgoing request


marcopolo39:
@microsoft-github-policy-service agree

